PR #3, opened 2019-11-22
feature: support t-head platform

xiaoxiacui:
1. Add T-HEAD ck802t procesoor
2. Add T-HEAD CB2201

### DIFF
--- a/FreeRTOS/Demo/T-HEAD_CB2201_GCC/FreeRTOSConfig.h
+++ b/FreeRTOS/Demo/T-HEAD_CB2201_GCC/FreeRTOSConfig.h
@@ -1,0 +1,128 @@
+/*
+    FreeRTOS V6.0.4 - Copyright (C) 2010 Real Time Engineers Ltd.
+
+    ***************************************************************************
+    *                                                                         *
+    * If you are:                                                             *
+    *                                                                         *
+    *    + New to FreeRTOS,                                                   *
+    *    + Wanting to learn FreeRTOS or multitasking in general quickly       *
+    *    + Looking for basic training,                                        *
+    *    + Wanting to improve your FreeRTOS skills and productivity           *
+    *                                                                         *
+    * then take a look at the FreeRTOS eBook                                  *
+    *                                                                         *
+    *        "Using the FreeRTOS Real Time Kernel - a Practical Guide"        *
+    *                  http://www.FreeRTOS.org/Documentation                  *
+    *                                                                         *
+    * A pdf reference manual is also available.  Both are usually delivered   *
+    * to your inbox within 20 minutes to two hours when purchased between 8am *
+    * and 8pm GMT (although please allow up to 24 hours in case of            *
+    * exceptional circumstances).  Thank you for your support!                *
+    *                                                                         *
+    ***************************************************************************
+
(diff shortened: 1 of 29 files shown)


RichardBarry:
Thanks for your submission.  A few questions and comments to start off:

- Do you have documentation that points people to the required development tools and development boards, and explains how to program the demo into the device along with the demo's behaviour (which is pretty standard with other demos)?

- Also, all the FreeRTOS kernel demos in the download can be built and debugged via an IDE, is that a possibility in this case?

- Are you ok with changing the comments at the top of the source files to the MIT license as per the other port files now?

- We will have to move the portable/GCC/T-HEAD_CK802T directory to portable/ThirdParty/GCC/T-HEAD_CK802T.

- Is the port specific to the CK802T, or can we use a more generic core name, such as CK80xT?  Sorry, not familiar with this family, but if the port is applicable to a wider family then we should name the directory to indicate all the chips it will work with.

- configASSERT() needs to be in the FreeRTOSConfig.h header file, rather than the portmacro.h, because it is something the application writer will need to define and undefined on an application by application basis (hence the definition is part of the application, rather than part of the port).

.......no doubt more will follow but I would like to be able to build the code myself first. 

Thanks again!

xiaoxiacui:
发件人: RichardBarry <notifications@github.com> 
发送时间: 2019年11月27日 8:19
收件人: FreeRTOS/FreeRTOS <FreeRTOS@noreply.github.com>
抄送: xiaoxiacui <xiaoxia_cui@c-sky.com>; Author <author@noreply.github.com>
主题: Re: [FreeRTOS/FreeRTOS] feature: support t-head platform (#3)

 

Thanks for your submission. A few questions and comments to start off:

*      Do you have documentation that points people to the required development tools and development boards, and explains how to program the demo into the device along with the demo's behaviour (which is pretty standard with other demos)?

[Vincent] yes, we have! 

*      Also, all the FreeRTOS kernel demos in the download can be built and debugged via an IDE, is that a possibility in this case?

[Vincent] there is two way for it. One is using command line to compile and download to run . another is using CDK IDE we developed by ourselves. But I suggest the command line method .  CDK IDE will be imported with RISC-V E902

*      Are you ok with changing the comments at the top of the source files to the MIT license as per the other port files now?

[Vincent] I need to check it with our adviser. We always to open source code with it before.

*      We will have to move the portable/GCC/T-HEAD_CK802T directory to portable/ThirdParty/GCC/T-HEAD_CK802T.

[Vincent] yes. It could be .

*      Is the port specific to the CK802T, or can we use a more generic core name, such as CK80xT? Sorry, not familiar with this family, but if the port is applicable to a wider family then we should name the directory to indicate all the chips it will work with.

[Vincent] in fact, the CK802 FAMILTY is with different features it could be T, S, F and other.  I will check it out and list all.

*      configASSERT() needs to be in the FreeRTOSConfig.h header file, rather than the portmacro.h, because it is something the application writer will need to define and undefined on an application by application basis (hence the definition is part of the application, rather than part of the port).

[Vincent] the code is built with v8.2.3 before. I am not sure it is compatible with the latest version.   But I found there are lot of other porting base on old version as well .

.......no doubt more will follow but I would like to be able to build the code myself first.

Thanks again!

PR #5, opened 2020-01-10
Fix inaccurate ticks in windows port

leegeth:
Root cause: Sleep() in Windows is not guaranteed to be realtime. Usage of Sleep()
was causing ticks in Windows port to be incorrect.
Solution: Use Waitable Timer in Windows for more accuracy.

### DIFF
--- a/FreeRTOS/Source/portable/MSVC-MingW/port.c
+++ b/FreeRTOS/Source/portable/MSVC-MingW/port.c
@@ -140,6 +140,9 @@ static DWORD WINAPI prvSimulatedPeripheralTimer( LPVOID lpParameter )
 {
 TickType_t xMinimumWindowsBlockTime;
 TIMECAPS xTimeCaps;
+TickType_t xWaitTimeBetweenTicks = portTICK_PERIOD_MS;
+HANDLE hTimer = NULL;
+LARGE_INTEGER liDueTime;
 
 	/* Set the timer resolution to the maximum possible. */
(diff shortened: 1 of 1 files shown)


RichardBarry:
This is a great change - although currently the change in timing makes some of our self monitoring tests fail as they were tuned to using the current method.  We will need to fix up the test timing at the same time as accepting this PR.

FreeRTOS/Source/portable/MSVC-MingW/port.c
```diff
@@ -159,22 +162,32 @@ TIMECAPS xTimeCaps;
...
+	liDueTime.u.HighPart = 0;
+
+	/* Create a synchronization Waitable Timer.*/
+	hTimer = CreateWaitableTimer( NULL, FALSE, NULL );
+
+	configASSERT( hTimer != NULL );
+
+	/* Set the Waitable Timer. The timer is set to run periodically at every
+	xWaitTimeBetweenTicks milliseconds. */
+	configASSERT( SetWaitableTimer( hTimer, &liDueTime, xWaitTimeBetweenTicks, NULL, NULL, 0 ) );
```
qiutongs, line 184: Nit-pick: better to separate to two lines for debugging purpose.

cobusve:
This one has become stale since the Kernel is now a submodule. I recreated it as https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/142

PR #6, opened 2020-02-11
Remove version number from kernel source files

yuhui-zheng:
The only directory touched in this commit is ```FreeRTOS/Source```. 

A second PR will remove version number in other directories.


yuhui-zheng:
This is mostly for dry run. Replacing with a better PR.

PR #7, opened 2020-02-11
Semantic versioning -- working progress. 

yuhui-zheng:
As subject, substitute "FreeRTOS Kernel V10.3.0" with "FreeRTOS Kernel" for all files under ```./FreeRTOS/Demo```.

###Related PR:
https://github.com/FreeRTOS/FreeRTOS/pull/6


RichardBarry:
Please don't leave the version number blank - if it is the head revision rather than a tagged version then use something like "tip revision" or "head revision".  Additionally it is important that the version string is identical in any file so we can update all the version numbers is one go simply by using an unambiguous string replacement script.

PR #22, opened 2020-02-27
CBMC kernel proof migration

yuhui-zheng:
This is a quick dirty migration to bring existing proofs to the new directory:
```./FreeRTOS/Test/CBMC```

Minimum modifications are done, including:
- path modification in scripts, config files and include directories.
- ```TaskPool/``` directory is renamed to ```Task/```, since it's not a kernel concept.
- Task proofs originally has dependency on TCP headers, which does not make too much sense. Thus removed.
- default ```FreeRTOSConfig.h``` is proposed under ```./patches``` directory. Unnecessary dependencies and defines are removed. This is not necessarily final.
- other minor stuffs. See commit messages.

This sequence should work: 
1. ```./prepare.py``` in ```./FreeRTOS/Test/CBMC/proofs/```
2. ```make report``` in each proof directory. 
(I'm skipping setup before step 1. Every proof under ./Task is tested before PR. Some proofs under ./Queue are spot checked.)

PR is raised to give folks a chance to review. Further clean-up shall be done after this PR is merged && git hook is setup. Clean up includes but not limited to: 
- patches directory contains all original patches, some of them can be removed.
- finalize ```FreeRTOSConfig.h``` and related make config in each proof.
- proof clean up and simplification.


yuhui-zheng:
See #23.

PR #25, opened 2020-03-05
use relative path to point to bundled toolchain instead

dachalco:
FreedomStudio comes with a  bundled installation of the riscv cross compiler tools. 
However our project had a hardcoded path to said bundled toolchain. 
Updated to relative point to IDEs bundled toolchain


dachalco:
FYI:
The setting change was made through the IDE, instead of XML

PR #24, opened 2020-02-27
Using relative url for submoduling.

yuhui-zheng:
Issue: https://github.com/FreeRTOS/FreeRTOS/issues/21

PR #27, opened 2020-03-06
Update readme and revert relative URL. 

yuhui-zheng:
As seen in commit message, three commits in order:
- Moving clone instruction up. 
- Rewording README.md to state clearly kernel is a submodule.
- Reverting relative URL change ( PR #24 ). (The functionality of relative URL is correct --  user could clone as expected. But the link is dead in GitHub UI, that user cannot click through. See the attached snapshot.)
 
Screen Shot 2020-03-05 at 11 15 46 PM

If any concern about the wording, please edit directly.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "FreeRTOS/Source"]
 	path = FreeRTOS/Source
-	url = ../FreeRTOS-Kernel.git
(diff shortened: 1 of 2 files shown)


chrisnc:
:( FWIW here is a thread about the Github issue: https://github.community/t5/How-to-use-Git-and-GitHub/Support-Linking-Relative-URL-s-on-submodules/td-p/24650

PR #31, opened 2020-03-09
logging about resources related to FreeRTOS+TCP

AniruddhaKanhere:
Enable a bit more logging to see what is actually happening

See [this](https://github.com/aws/amazon-freertos/pull/1626) PR for rationale and discussion. This PR adds certain more changes but since FreeRTOS doesn't have any directory such as '/vendors', I haven't added those changes here. I am just adding the change necessary for logging.


FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/portable/NetworkInterface/esp32/NetworkInterface.c
```diff
@@ -44,6 +44,10 @@ enum if_state_t {
 static const char *TAG = "NetInterface";
 volatile static uint32_t xInterfaceState = INTERFACE_DOWN;
 
+#if ( ipconfigHAS_PRINTF != 0 )
```
cobusve: What else switches on when you have this defined? It seems to me like we need to be able to switch off this monitoring printouts and keep the Printf definition for other purposes?

RichardBarry: Would be better to make prvMonitorResources() a macro that only exists when ipconfigHAS_PRINTF is defined as != 0 rather than guard each call to prvMonitorResources() with its own #if block.

cobusve: Yes I agree, less opportunity to miss something.

AniruddhaKanhere: Agreed.
But this (copying of one file at a time) is not a good was to achieve SSoT. Let this PR stay open until something is decided. Then if necessary, I will make the changes

PR #39, opened 2020-03-19
Fix function comments

maramon:
Fix function comments

Description
-----------
This change fixes incorrect function comments.

Test Steps
-----------
N/A

Related Issue
-----------
N/A


yuhui-zheng:
Hi @maramon 

Thank you for submitting this PR. 

The driver code is usually from MCU vendor distribution, that changing it

PR #46, opened 2020-04-01
ARP: Coverity + MISRA compliance

AniruddhaKanhere:
This PR focuses on resolving the MISRA compliance related warnings in FreeRTOS_ARP.c.
This doesn't ensure that it doesn't have any warnings at all since a lot of warnigns require code restructure/rework and some warnings require editing header files and some are portable layer warnings.

Test Steps
-----------
Ran all Microchip Curiosity Tests on CI. Thereby ensuring buildability and executability of the code. GGD tests failed.

### DIFF
--- a/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
+++ b/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
@@ -85,9 +85,16 @@ static TickType_t xLastGratuitousARPTime = ( TickType_t ) 0;
 	MACAddress_t xARPClashMacAddress;
 #endif /* ipconfigARP_USE_CLASH_DETECTION */
 
+
 /* Part of the Ethernet and ARP headers are always constant when sending an IPv4
 ARP packet.  This array defines the constant parts, allowing this part of the
 packet to be filled in using a simple memcpy() instead of individual writes. */
+
+/* MISRA c2012 rule 8.9 violation: "xDefaultPartARPPacketHeader" should be defined at block
+ * scope.
+ * Rule has been relaxed since this is a configuration parameter. And this location provides
+ * more readability.
+ */
 static const uint8_t xDefaultPartARPPacketHeader[] =
 {
 	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 	/* Ethernet destination address. */
@@ -114,7 +121,7 @@ uint32_t ulTargetProtocolAddress, ulSenderProtocolAddress;
 	pxARPHeader = &( pxARPFrame->xARPHeader );
 
 	/* The field ulSenderProtocolAddress is badly aligned, copy byte-by-byte. */
-	memcpy( ( void *)&( ulSenderProtocolAddress ), ( void * )pxARPHeader->ucSenderProtocolAddress, sizeof( ulSenderProtocolAddress ) );
+	( void ) memcpy( ( void *)&( ulSenderProtocolAddress ), ( void * )pxARPHeader->ucSenderProtocolAddress, sizeof( ulSenderProtocolAddress ) );
 	/* The field ulTargetProtocolAddress is well-aligned, a 32-bits copy. */
 	ulTargetProtocolAddress = pxARPHeader->ulTargetProtocolAddress;
 
@@ -144,17 +151,17 @@ uint32_t ulTargetProtocolAddress, ulSenderProtocolAddress;
 					{
 						/* A double IP address is detected! */
 						/* Give the sources MAC address the value of the broadcast address, will be swapped later */
-						memcpy( pxARPFrame->xEthernetHeader.xSourceAddress.ucBytes, xBroadcastMACAddress.ucBytes, sizeof( xBroadcastMACAddress ) );
-						memset( pxARPHeader->xTargetHardwareAddress.ucBytes, '\0', sizeof( MACAddress_t ) );
+						( void ) memcpy( pxARPFrame->xEthernetHeader.xSourceAddress.ucBytes, xBroadcastMACAddress.ucBytes, sizeof( xBroadcastMACAddress ) );
+						( void ) memset( pxARPHeader->xTargetHardwareAddress.ucBytes, (int)'\0', sizeof( MACAddress_t ) );
 						pxARPHeader->ulTargetProtocolAddress = 0UL;
 					}
 					else
 					{
-						memcpy( pxARPHeader->xTargetHardwareAddress.ucBytes, pxARPHeader->xSenderHardwareAddress.ucBytes, sizeof( MACAddress_t ) );
+						( void ) memcpy( pxARPHeader->xTargetHardwareAddress.ucBytes, pxARPHeader->xSenderHardwareAddress.ucBytes, sizeof( MACAddress_t ) );
 						pxARPHeader->ulTargetProtocolAddress = ulSenderProtocolAddress;
 					}
-					memcpy( pxARPHeader->xSenderHardwareAddress.ucBytes, ( void * ) ipLOCAL_MAC_ADDRESS, sizeof( MACAddress_t ) );
-					memcpy( ( void* )pxARPHeader->ucSenderProtocolAddress, ( void* )ipLOCAL_IP_ADDRESS_POINTER, sizeof( pxARPHeader->ucSenderProtocolAddress ) );
+					( void ) memcpy( pxARPHeader->xSenderHardwareAddress.ucBytes, ( void * ) ipLOCAL_MAC_ADDRESS, sizeof( MACAddress_t ) );
+					( void ) memcpy( ( void* )pxARPHeader->ucSenderProtocolAddress, ( void* )ipLOCAL_IP_ADDRESS_POINTER, sizeof( pxARPHeader->ucSenderProtocolAddress ) );
 
 					eReturn = eReturnEthernetFrame;
 				}
@@ -169,7 +176,7 @@ uint32_t ulTargetProtocolAddress, ulSenderProtocolAddress;
 					if( ulSenderProtocolAddress == *ipLOCAL_IP_ADDRESS_POINTER )
 					{
 						xARPHadIPClash = pdTRUE;
-						memcpy( xARPClashMacAddress.ucBytes, pxARPHeader->xSenderHardwareAddress.ucBytes, sizeof( xARPClashMacAddress.ucBytes ) );
+						( void ) memcpy( xARPClashMacAddress.ucBytes, pxARPHeader->xSenderHardwareAddress.ucBytes, sizeof( xARPClashMacAddress.ucBytes ) );
 					}
 				}
 				#endif /* ipconfigARP_USE_CLASH_DETECTION */
@@ -198,7 +205,7 @@ uint32_t ulTargetProtocolAddress, ulSenderProtocolAddress;
 			if( ( memcmp( xARPCache[ x ].xMACAddress.ucBytes, pxMACAddress->ucBytes, sizeof( pxMACAddress->ucBytes ) ) == 0 ) )
 			{
 				lResult = xARPCache[ x ].ulIPAddress;
-				memset( &xARPCache[ x ], '\0', sizeof( xARPCache[ x ] ) );
+				( void ) memset( &xARPCache[ x ], '\0', sizeof( xARPCache[ x ] ) );
 				break;
 			}
 		}
@@ -222,7 +229,7 @@ uint8_t ucMinAgeFound = 0U;
 		Unless: when '*ipLOCAL_IP_ADDRESS_POINTER' equals zero, the IP-address
 		and netmask are still unknown. */
 		if( ( ( ulIPAddress & xNetworkAddressing.ulNetMask ) == ( ( *ipLOCAL_IP_ADDRESS_POINTER ) & xNetworkAddressing.ulNetMask ) ) ||
-			( *ipLOCAL_IP_ADDRESS_POINTER == 0ul ) )
+			( *ipLOCAL_IP_ADDRESS_POINTER == 0UL ) )
 	#else
 		/* If ipconfigARP_STORES_REMOTE_ADDRESSES is non-zero, IP addresses with
 		a different netmask will also be stored.  After when replying to a UDP
@@ -261,6 +268,8 @@ uint8_t ucMinAgeFound = 0U;
 					optimisation. */
 					xARPCache[ x ].ucAge = ( uint8_t ) ipconfigMAX_ARP_AGE;
 					xARPCache[ x ].ucValid = ( uint8_t ) pdTRUE;
+
+					/* MISRA rule 15.5 relaxed for readability of code */
 					return;
 				}
 
@@ -302,6 +311,10 @@ uint8_t ucMinAgeFound = 0U;
 				ucMinAgeFound = xARPCache[ x ].ucAge;
 				xUseEntry = x;
 			}
+			else    /* Added to supress MISRA rule 15.7 voilation */
+			{
+				/* Do nothing */
+			}
 		}
 
 		if( xMacEntry >= 0 )
@@ -313,7 +326,8 @@ uint8_t ucMinAgeFound = 0U;
 				/* Both the MAC address as well as the IP address were found in
 				different locations: clear the entry which matches the
 				IP-address */
-				memset( &xARPCache[ xIpEntry ], '\0', sizeof( xARPCache[ xIpEntry ] ) );
+				/* Silence MISRA warning about unused return value */
+				( void ) memset( &xARPCache[ xIpEntry ], (int)'\0', sizeof( xARPCache[ xIpEntry ] ) );
 			}
 		}
 		else if( xIpEntry >= 0 )
@@ -321,13 +335,17 @@ uint8_t ucMinAgeFound = 0U;
 			/* An entry containing the IP-address was found, but it had a different MAC address */
 			xUseEntry = xIpEntry;
 		}
+		else
+		{
+			/* Do nothing! xUseEntry retains its value */
+		}
 
 		/* If the entry was not found, we use the oldest entry and set the IPaddress */
 		xARPCache[ xUseEntry ].ulIPAddress = ulIPAddress;
 
 		if( pxMACAddress != NULL )
 		{
-			memcpy( xARPCache[ xUseEntry ].xMACAddress.ucBytes, pxMACAddress->ucBytes, sizeof( pxMACAddress->ucBytes ) );
+			( void ) memcpy( xARPCache[ xUseEntry ].xMACAddress.ucBytes, pxMACAddress->ucBytes, sizeof( pxMACAddress->ucBytes ) );
 
 			iptraceARP_TABLE_ENTRY_CREATED( ulIPAddress, (*pxMACAddress) );
 			/* And this entry does not need immediate attention */
@@ -338,6 +356,10 @@ uint8_t ucMinAgeFound = 0U;
 		{
 			xARPCache[ xUseEntry ].ucAge = ( uint8_t ) ipconfigMAX_ARP_RETRANSMISSIONS;
 			xARPCache[ xUseEntry ].ucValid = ( uint8_t ) pdFALSE;
+		}
+		else
+		{
+			/* Control should not get here */
 		}
 	}
 }
@@ -377,7 +399,7 @@ uint32_t ulAddressToLookup;
 	if( *pulIPAddress == ipLLMNR_IP_ADDR )	/* Is in network byte order. */
 	{
 		/* The LLMNR IP-address has a fixed virtual MAC address. */
-		memcpy( pxMACAddress->ucBytes, xLLMNR_MacAdress.ucBytes, sizeof( MACAddress_t ) );
+		( void ) memcpy( pxMACAddress->ucBytes, xLLMNR_MacAdress.ucBytes, sizeof( MACAddress_t ) );
 		eReturn = eARPCacheHit;
 	}
 	else
@@ -386,7 +408,7 @@ uint32_t ulAddressToLookup;
 		( *pulIPAddress == xNetworkAddressing.ulBroadcastAddress ) )/* Or a local broadcast address, eg 192.168.1.255? */
 	{
 		/* This is a broadcast so uses the broadcast MAC address. */
-		memcpy( pxMACAddress->ucBytes, xBroadcastMACAddress.ucBytes, sizeof( MACAddress_t ) );
+		( void ) memcpy( pxMACAddress->ucBytes, xBroadcastMACAddress.ucBytes, sizeof( MACAddress_t ) );
 		eReturn = eARPCacheHit;
 	}
 	else if( *ipLOCAL_IP_ADDRESS_POINTER == 0UL )
@@ -479,7 +501,7 @@ eARPLookupResult_t eReturn = eARPCacheMiss;
 			else
 			{
 				/* A valid entry was found. */
-				memcpy( pxMACAddress->ucBytes, xARPCache[ x ].xMACAddress.ucBytes, sizeof( MACAddress_t ) );
+				( void ) memcpy( pxMACAddress->ucBytes, xARPCache[ x ].xMACAddress.ucBytes, sizeof( MACAddress_t ) );
 				eReturn = eARPCacheHit;
 			}
 			break;
@@ -549,7 +571,7 @@ void vARPSendGratuitous( void )
 	xLastGratuitousARPTime = ( TickType_t ) 0;
 
 	/* Let the IP-task call vARPAgeCache(). */
-	xSendEventToIPTask( eARPTimerEvent );
+	( void ) xSendEventToIPTask( eARPTimerEvent );
 }
 
 /*-----------------------------------------------------------*/
@@ -583,7 +605,7 @@ NetworkBufferDescriptor_t *pxNetworkBuffer;
 		if( xIsCallingFromIPTask() != 0 )
 		{
 			/* Only the IP-task is allowed to call this function directly. */
-			xNetworkInterfaceOutput( pxNetworkBuffer, pdTRUE );
+			( void ) xNetworkInterfaceOutput( pxNetworkBuffer, pdTRUE );
 		}
 		else
 		{
@@ -608,7 +630,7 @@ ARPPacket_t *pxARPPacket;
 	/* Buffer allocation ensures that buffers always have space
 	for an ARP packet. See buffer allocation implementations 1
 	and 2 under portable/BufferManagement. */
-	configASSERT( pxNetworkBuffer );
+	configASSERT( pxNetworkBuffer != NULL );
 	configASSERT( pxNetworkBuffer->xDataLength >= sizeof(ARPPacket_t) );
 
 	pxARPPacket = ( ARPPacket_t * ) pxNetworkBuffer->pucEthernetBuffer;
@@ -624,11 +646,16 @@ ARPPacket_t *pxARPPacket;
 		xARPHeader.usOperation;
 		xARPHeader.xTargetHardwareAddress;
 	*/
-	memcpy( ( void * ) pxARPPacket, ( void * ) xDefaultPartARPPacketHeader, sizeof( xDefaultPartARPPacketHeader ) );
-	memcpy( ( void * ) pxARPPacket->xEthernetHeader.xSourceAddress.ucBytes , ( void * ) ipLOCAL_MAC_ADDRESS, ( size_t ) ipMAC_ADDRESS_LENGTH_BYTES );
-	memcpy( ( void * ) pxARPPacket->xARPHeader.xSenderHardwareAddress.ucBytes, ( void * ) ipLOCAL_MAC_ADDRESS, ( size_t ) ipMAC_ADDRESS_LENGTH_BYTES );
 
-	memcpy( ( void* )pxARPPacket->xARPHeader.ucSenderProtocolAddress, ( void* )ipLOCAL_IP_ADDRESS_POINTER, sizeof( pxARPPacket->xARPHeader.ucSenderProtocolAddress ) );
+	/* Casting xDefaultPartARPPacketHeader as ("const" void*). Else it would 
+         * be a MISRA c2012 rule 11.18 violation.
+	 * Also, for rule 21.15 regarding using same pointer-to-x types for memcpy,
+         * below is done intentionally here and thus the rule is relaxed         */
+	( void ) memcpy( ( void * ) pxARPPacket, (const void * ) xDefaultPartARPPacketHeader, sizeof( xDefaultPartARPPacketHeader ) );
+	( void ) memcpy( ( void * ) pxARPPacket->xEthernetHeader.xSourceAddress.ucBytes , ( void * ) ipLOCAL_MAC_ADDRESS, ( size_t ) ipMAC_ADDRESS_LENGTH_BYTES );
+	( void ) memcpy( ( void * ) pxARPPacket->xARPHeader.xSenderHardwareAddress.ucBytes, ( void * ) ipLOCAL_MAC_ADDRESS, ( size_t ) ipMAC_ADDRESS_LENGTH_BYTES );
+
+	( void ) memcpy( ( void* )pxARPPacket->xARPHeader.ucSenderProtocolAddress, ( void* )ipLOCAL_IP_ADDRESS_POINTER, sizeof( pxARPPacket->xARPHeader.ucSenderProtocolAddress ) );
 	pxARPPacket->xARPHeader.ulTargetProtocolAddress = pxNetworkBuffer->ulIPAddress;
 
 	pxNetworkBuffer->xDataLength = sizeof( ARPPacket_t );
@@ -639,12 +666,18 @@ ARPPacket_t *pxARPPacket;
 
 void FreeRTOS_ClearARP( void )
 {
-	memset( xARPCache, '\0', sizeof( xARPCache ) );
+	/* Function requires the second argument to be an int.
+	 *
+	 * MISRA rule 17.7 relaxed. The return value is not 
+	 * required to be used. Also improves readability    */
+	( void ) memset( xARPCache, (int)'\0', sizeof( xARPCache ) );
 }
 /*-----------------------------------------------------------*/
 
 #if( ipconfigHAS_PRINTF != 0 ) || ( ipconfigHAS_DEBUG_PRINTF != 0 )
 
+	/* MISRA c 2012 rule 8.7 relaxed since this function can be
+	 * called from external location when debugging is enabled */
 	void FreeRTOS_PrintARPCache( void )
 	{
 	BaseType_t x, xCount = 0;
@@ -652,7 +685,7 @@ void FreeRTOS_ClearARP( void )
 		/* Loop through each entry in the ARP cache. */
 		for( x = 0; x < ipconfigARP_CACHE_ENTRIES; x++ )
 		{
-			if( ( xARPCache[ x ].ulIPAddress != 0ul ) && ( xARPCache[ x ].ucAge > 0U ) )
+			if( ( xARPCache[ x ].ulIPAddress != 0UL ) && ( xARPCache[ x ].ucAge > 0U ) )
 			{
 				/* See if the MAC-address also matches, and we're all happy */
 				FreeRTOS_printf( ( "Arp %2ld: %3u - %16lxip : %02x:%02x:%02x : %02x:%02x:%02x\n",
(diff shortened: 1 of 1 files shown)


FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -583,7 +605,7 @@ NetworkBufferDescriptor_t *pxNetworkBuffer;
 		if( xIsCallingFromIPTask() != 0 )
 		{
 			/* Only the IP-task is allowed to call this function directly. */
-			xNetworkInterfaceOutput( pxNetworkBuffer, pdTRUE );
+			( void ) xNetworkInterfaceOutput( pxNetworkBuffer, pdTRUE );
```
yuhui-zheng, line 608: Is it possible to check the return value here?

```xNetworkInterfaceOutput``` is implemented by third party, that I would check the return value to be sure the API contract is met.

File: FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c [hunk 2]
```diff
@@ -144,17 +151,17 @@ uint32_t ulTargetProtocolAddress, ulSenderProtocolAddress;
 					{
 						/* A double IP address is detected! */
 						/* Give the sources MAC address the value of the broadcast address, will be swapped later */
-						memcpy( pxARPFrame->xEthernetHeader.xSourceAddress.ucBytes, xBroadcastMACAddress.ucBytes, sizeof( xBroadcastMACAddress ) );
-						memset( pxARPHeader->xTargetHardwareAddress.ucBytes, '\0', sizeof( MACAddress_t ) );
+						( void ) memcpy( pxARPFrame->xEthernetHeader.xSourceAddress.ucBytes, xBroadcastMACAddress.ucBytes, sizeof( xBroadcastMACAddress ) );
+						( void ) memset( pxARPHeader->xTargetHardwareAddress.ucBytes, (int)'\0', sizeof( MACAddress_t ) );
```
yuhui-zheng, line 155: ```memset``` at this line and at 208 are inconsistent. Recommending to stick with a pattern -- either cast or just 0.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -639,20 +666,26 @@ ARPPacket_t *pxARPPacket;
...
+	 *
+	 * MISRA rule 17.7 relaxed. The return value is not 
+	 * required to be used. Also improves readability    */
+	( void ) memset( xARPCache, (int)'\0', sizeof( xARPCache ) );
 }
 /*-----------------------------------------------------------*/
 
 #if( ipconfigHAS_PRINTF != 0 ) || ( ipconfigHAS_DEBUG_PRINTF != 0 )
 
+	/* MISRA c 2012 rule 8.7 relaxed since this function can be
```
yuhui-zheng, line 679: Is the rule "relaxed" here? 

MISRA 2012 8.7 -- https://wiki.sei.cmu.edu/confluence/pages/viewpage.action?pageId=87152278. 
```FreeRTOS_PrintARPCache``` is an API, as defined in FreeRTOS_IP.h. Maybe I'm missing something?

File: FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c (hunk 1 above)
Comment (AniruddhaKanhere) [reply FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c:608]: There is no way to return the error to the caller since the function returns a void. I can go ahead and add a configASSERT which seems drastic to me.
Or, as you suggested I can add a logging instead of 'configASSERT'ing.
@htibosch, any comments on what would be better?

File: FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c (hunk 3 above)
Comment (AniruddhaKanhere) [reply FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c:679]: The sole purpose of function `FreeRTOS_PrintARPCache` is to be used for debugging. It can be called by a user for debugging the code. MISRA rule 8.7 requires that I declare this function as static (since this function is not being used in this code base).  If I declare this function as static, users will not be able to use this function.

File: FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c (hunk 1 above)
Comment (htibosch) [reply FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c:608]: Hello, the function `xNetworkInterfaceOutput()` is supposed to return a non-zero value if it was successful. The return value could be logged, but the is no different behaviour of the IP-task: when a packet can not be delivered, that is just bad luck. it will be tried later.

Comment (AniruddhaKanhere) [reply FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c]: Thanks Hein, I have added a looging functionality for easier debugging.

File: FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c (hunk 2 above)
Comment (AniruddhaKanhere) [reply FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c]: done. thanks 😄

File: FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c [hunk 4]
```diff
@@ -302,6 +311,10 @@ uint8_t ucMinAgeFound = 0U;
 				ucMinAgeFound = xARPCache[ x ].ucAge;
 				xUseEntry = x;
 			}
+			else    /* Added to supress MISRA rule 15.7 voilation */
```
gkwicker, line 314: spelling: _violation_

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -261,6 +268,8 @@ uint8_t ucMinAgeFound = 0U;
 					optimisation. */
 					xARPCache[ x ].ucAge = ( uint8_t ) ipconfigMAX_ARP_AGE;
 					xARPCache[ x ].ucValid = ( uint8_t ) pdTRUE;
+
+					/* MISRA rule 15.5 relaxed for readability of code */
```
gkwicker, line 272: How are you suppressing this violation?  I don't see a coverity annotation here.  Also, could this code be refactored?  For example, I think we'll also bump into the complexity > 8 issue in this function, it has a lot of branching.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -313,21 +326,26 @@ uint8_t ucMinAgeFound = 0U;
 				/* Both the MAC address as well as the IP address were found in
 				different locations: clear the entry which matches the
 				IP-address */
-				memset( &xARPCache[ xIpEntry ], '\0', sizeof( xARPCache[ xIpEntry ] ) );
+				/* Silence MISRA warning about unused return value */
```
gkwicker: Delete this comment.  You're casting to `void` here because you're not interested in the return value (i.e. it's the right thing to do), not because you want to get rid of a MISRA warning.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -339,6 +357,10 @@ uint8_t ucMinAgeFound = 0U;
 			xARPCache[ xUseEntry ].ucAge = ( uint8_t ) ipconfigMAX_ARP_RETRANSMISSIONS;
 			xARPCache[ xUseEntry ].ucValid = ( uint8_t ) pdFALSE;
 		}
+		else
+		{
+			/* Control should not get here */
```
gkwicker, line 361: Out of scope for this PR – but what if it does?

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
gkwicker: The best thing to do would be to count these and have some utility for retrieving counts for intermittent logging (say once per minute or via CLI command).  Having counts of delivered/undelivered packets, etc. is incredibly useful for debugging and can be done without affecting runtime behavior as happens with logging.

However, this is out of scope for this PR.  The purpose of this PR is not to add new capabilities to the code, but to eliminate some Coverity warnings.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -624,11 +646,16 @@ ARPPacket_t *pxARPPacket;
 		xARPHeader.usOperation;
 		xARPHeader.xTargetHardwareAddress;
 	*/
-	memcpy( ( void * ) pxARPPacket, ( void * ) xDefaultPartARPPacketHeader, sizeof( xDefaultPartARPPacketHeader ) );
-	memcpy( ( void * ) pxARPPacket->xEthernetHeader.xSourceAddress.ucBytes , ( void * ) ipLOCAL_MAC_ADDRESS, ( size_t ) ipMAC_ADDRESS_LENGTH_BYTES );
-	memcpy( ( void * ) pxARPPacket->xARPHeader.xSenderHardwareAddress.ucBytes, ( void * ) ipLOCAL_MAC_ADDRESS, ( size_t ) ipMAC_ADDRESS_LENGTH_BYTES );
 
-	memcpy( ( void* )pxARPPacket->xARPHeader.ucSenderProtocolAddress, ( void* )ipLOCAL_IP_ADDRESS_POINTER, sizeof( pxARPPacket->xARPHeader.ucSenderProtocolAddress ) );
+	/* Casting xDefaultPartARPPacketHeader as ("const" void*). Else it would
```
gkwicker, line 653: Delete this comment.  You are casting to `const void *` because that is what the prototype for `memcpy()` expects, i.e. it is the right thing to do.  That Coverity no longer tells you there is a warning here merely confirms that you have done the right thing :-)

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -639,20 +666,26 @@ ARPPacket_t *pxARPPacket;
 
 void FreeRTOS_ClearARP( void )
 {
-	memset( xARPCache, '\0', sizeof( xARPCache ) );
+	/* Function requires the second argument to be an int.
+	 *
+	 * MISRA rule 17.7 relaxed. The return value is not 
+	 * required to be used. Also improves readability    */
+	( void ) memset( xARPCache, (int)'\0', sizeof( xARPCache ) );
```
gkwicker: Delete comment on lines 669-672 and replace with:

```c
 	( void ) memset( xARPCache, 0, sizeof( xARPCache ) ); 
```

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
gkwicker: What does the comment on line 679 mean?  How is this rule relaxed here?  I don't see a Coverity annotation to disable it; if I run Coverity on this code will it still emit a warning here?

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
AniruddhaKanhere: I agree completely with @gkwicker

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
AniruddhaKanhere: done

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
RichardBarry: Please don't remove them - they are necessary to demonstrate you are intentionally uninterested in the return value - and is the only way of demonstrating that to an automated tool that won't read the comments.

gkwicker: +1 on this comment from @RichardBarry.  Every use of a function should be according to its prototype; if that function returns something and you are not interested in it, you should cast to `void`.

AniruddhaKanhere: Sure thing. I added back the (void) to all instances of memmove/memcpy etc.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
gkwicker: Reducing the branching will have to be done but you can do that in a separate PR.

AniruddhaKanhere: Sure thing 👍

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -114,7 +118,7 @@ uint32_t ulTargetProtocolAddress, ulSenderProtocolAddress;
 	pxARPHeader = &( pxARPFrame->xARPHeader );
 
 	/* The field ulSenderProtocolAddress is badly aligned, copy byte-by-byte. */
-	memcpy( ( void *)&( ulSenderProtocolAddress ), ( void * )pxARPHeader->ucSenderProtocolAddress, sizeof( ulSenderProtocolAddress ) );
+	( void ) memcpy( ( void * )&( ulSenderProtocolAddress ), ( void * )pxARPHeader->ucSenderProtocolAddress, sizeof( ulSenderProtocolAddress ) );
```
gkwicker, line 121: Second argument should be cast to `( const void * )`, `memcpy()` prototype is:

```c
       void *memcpy(void *dest, const void *src, size_t n);
```

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -187,7 +191,7 @@ uint32_t ulTargetProtocolAddress, ulSenderProtocolAddress;
 
 #if( ipconfigUSE_ARP_REMOVE_ENTRY != 0 )
 
-	uint32_t ulARPRemoveCacheEntryByMac( const MACAddress_t * pxMACAddress )
+	uint32_t ulARPRemoveCacheEntryByMac(const MACAddress_t* pxMACAddress)
```
gkwicker, line 194: replace the spaces from the previous version

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -261,6 +265,11 @@ uint8_t ucMinAgeFound = 0U;
 					optimisation. */
 					xARPCache[ x ].ucAge = ( uint8_t ) ipconfigMAX_ARP_AGE;
 					xARPCache[ x ].ucValid = ( uint8_t ) pdTRUE;
+
+					/* MISRA rule 15.5 relaxed for reduced complexity. Not putting
+					 * a return here will increase cyclomatic complexity in later
+					 * part of the code */
+					/* coverity[misra_c_2012_rule_15_5_violation] */
```
gkwicker, line 272: Do __not__ disable this rule and delete lines 268-272.  We __want__ Coverity to continue warning about this so that we remember to go back and refactor this function later.  The goal is __not__ to eliminate MISRA warnings, the goal is _to do the right thing._  And overly complex functions should be refactored.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -302,6 +311,10 @@ uint8_t ucMinAgeFound = 0U;
 				ucMinAgeFound = xARPCache[ x ].ucAge;
 				xUseEntry = x;
 			}
+			else    /* Added to suppress MISRA rule 15.7 violation */
```
gkwicker, line 314: You are not _suppressing_ a rule here, you are _conforming_ to it.  Delete the comment after the `else`.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -339,6 +356,12 @@ uint8_t ucMinAgeFound = 0U;
 			xARPCache[ xUseEntry ].ucAge = ( uint8_t ) ipconfigMAX_ARP_RETRANSMISSIONS;
 			xARPCache[ xUseEntry ].ucValid = ( uint8_t ) pdFALSE;
 		}
+		else
+		{
+			/* Do nothing. (pxMACAddress = NULL) and (xIPEntry > 0)
```
gkwicker, line 361: Correct to `( pxMACAddress == NULL )`

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -354,7 +377,7 @@ uint8_t ucMinAgeFound = 0U;
 		{
 			/* Does this row in the ARP cache table hold an entry for the MAC
 			address being searched? */
-			if( memcmp( pxMACAddress->ucBytes, xARPCache[ x ].xMACAddress.ucBytes, sizeof( MACAddress_t ) ) == 0 )
+			if( ( void ) memcmp( pxMACAddress->ucBytes, xARPCache[ x ].xMACAddress.ucBytes, sizeof( MACAddress_t ) ) == 0 )
```
gkwicker, line 380: Why are you casting the result of `memcmp()` away when you are using it in a comparison at the end of the statement (` == 0`)?  Delete the cast to `( void )`.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -624,11 +651,15 @@ ARPPacket_t *pxARPPacket;
...
 		xARPHeader.xTargetHardwareAddress;
 	*/
-	memcpy( ( void * ) pxARPPacket, ( void * ) xDefaultPartARPPacketHeader, sizeof( xDefaultPartARPPacketHeader ) );
-	memcpy( ( void * ) pxARPPacket->xEthernetHeader.xSourceAddress.ucBytes , ( void * ) ipLOCAL_MAC_ADDRESS, ( size_t ) ipMAC_ADDRESS_LENGTH_BYTES );
-	memcpy( ( void * ) pxARPPacket->xARPHeader.xSenderHardwareAddress.ucBytes, ( void * ) ipLOCAL_MAC_ADDRESS, ( size_t ) ipMAC_ADDRESS_LENGTH_BYTES );
 
-	memcpy( ( void* )pxARPPacket->xARPHeader.ucSenderProtocolAddress, ( void* )ipLOCAL_IP_ADDRESS_POINTER, sizeof( pxARPPacket->xARPHeader.ucSenderProtocolAddress ) );
+	/* Also, for rule 21.15 regarding using same pointer-to-x types for memcpy,
+	 * below is done intentionally here and thus the rule is relaxed         */
+	/* coverity[misra_c_2012_rule_21_15_violation] */
```
gkwicker, line 657: I don't believe you need to suppress this here.  Note that lines 659/660/662 do not conform to the prototype for `memcpy()` – the second argument should be cast to `( const void * ).  Correct that and you should be able tyo delete lines 655-657.

```c
       void *memcpy(void *dest, const void *src, size_t n);
```

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -88,6 +88,10 @@ static TickType_t xLastGratuitousARPTime = ( TickType_t ) 0;
 /* Part of the Ethernet and ARP headers are always constant when sending an IPv4
 ARP packet.  This array defines the constant parts, allowing this part of the
 packet to be filled in using a simple memcpy() instead of individual writes. */
+
+/* Rule has been relaxed since this is a configuration parameter. And this location provides
+ * more readability. */
```
yuhui-zheng, line 93: Assuming we are done with functional review, planning to glance over cosmetics. 

Minor: please keep comment block style consistent. This is not a blocker. 
```
/* line 1
line 2
line 3
*/
```
or 
```
/* line 1
 * line 2
 * line 3
 */
```

And, have a single block. So preferably _not_ 
```
statements...
/* don't */
/* do */
/* this */
statements...
```

```
statement...
/* Yes,
this looks better. */
statement...
```

yuhui-zheng: Since we have multiple PRs just for Coverity, for all the styling issues we could do a PR for entire +TCP after everything is in. And if that's the rule we all agree on (please check with all stakeholders), suggesting to leave out all style changes to make review easier. 

I'm not saying style is not important. I think fixing Coverity violations is more crucial. And also suggesting to get CI team involved to setup style checks.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -198,7 +202,7 @@ uint32_t ulTargetProtocolAddress, ulSenderProtocolAddress;
 			if( ( memcmp( xARPCache[ x ].xMACAddress.ucBytes, pxMACAddress->ucBytes, sizeof( pxMACAddress->ucBytes ) ) == 0 ) )
 			{
 				lResult = xARPCache[ x ].ulIPAddress;
-				memset( &xARPCache[ x ], '\0', sizeof( xARPCache[ x ] ) );
+				( void ) memset( &xARPCache[ x ], '\0', sizeof( xARPCache[ x ] ) );
```
yuhui-zheng, line 205: Please stick to the same style -- second argument 0.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
yuhui-zheng: +1. Casting breaks the logic, please fix.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
AniruddhaKanhere: With the latest PR, I added consistent comment style.
I am unsure about writting a big block of comment in case I am trying to suppress Coverity.
I believe I should write /* coverity[misra_related_warning_violation] */ to suppress the warnings.

htibosch: About coding style: all FreeRTOS code used to have this style:

~~~c
/* line 1
line 2
line 3 */
~~~

Currently, I see that this style is also used in `tasks.c`:

~~~c
/*
 * line 1
 * line 2
 * line 3
 */
~~~
So things might be changing...

Personally I like the second style

What is important is to start a comment with a capital letter, and end it with a period.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -114,47 +118,47 @@ uint32_t ulTargetProtocolAddress, ulSenderProtocolAddress;
...
-					pxARPHeader->usOperation = ( uint16_t ) ipARP_REPLY;
+					pxARPHeader->usOperation = ( uint16_t )ipARP_REPLY;
 					if( ulTargetProtocolAddress == ulSenderProtocolAddress )
 					{
 						/* A double IP address is detected! */
 						/* Give the sources MAC address the value of the broadcast address, will be swapped later */
-						memcpy( pxARPFrame->xEthernetHeader.xSourceAddress.ucBytes, xBroadcastMACAddress.ucBytes, sizeof( xBroadcastMACAddress ) );
-						memset( pxARPHeader->xTargetHardwareAddress.ucBytes, '\0', sizeof( MACAddress_t ) );
+						( void ) memcpy( ( void * ) pxARPFrame->xEthernetHeader.xSourceAddress.ucBytes, ( const void * ) xBroadcastMACAddress.ucBytes, sizeof( xBroadcastMACAddress ) );
+						( void ) memset( pxARPHeader->xTargetHardwareAddress.ucBytes, 0, sizeof( MACAddress_t ) );
```
gkwicker, line 152: Why no `( const void * )` on argument 2 here?  You have it everywhere else.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -583,7 +601,11 @@ NetworkBufferDescriptor_t *pxNetworkBuffer;
 		if( xIsCallingFromIPTask() != 0 )
 		{
 			/* Only the IP-task is allowed to call this function directly. */
-			xNetworkInterfaceOutput( pxNetworkBuffer, pdTRUE );
+			if( xNetworkInterfaceOutput( pxNetworkBuffer, pdTRUE ) != pdTRUE )
+			{
+				/* Not sent? Bad news. Maybe link is down? */
+				FreeRTOS_printf( ( "xNetworkInterfaceOutput failed. Link down?\n" ) );
```
gkwicker, line 607: I don't know if we want to use a general `FreeRTOS_printf()` here, since this could produce a lot of console output in normal operation.  Since we don't have a network interface metrics object (that I am aware of), for now, let's just make sure that debugging output has been configured in the build.  See [example here.](https://github.com/FreeRTOS/FreeRTOS/blob/master/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c#L646)

```c
if( xNetworkInterfaceOutput( pxNetworkBuffer, pdTRUE ) != pdTRUE )
{
#if( ipconfigHAS_PRINTF != 0 ) || ( ipconfigHAS_DEBUG_PRINTF != 0 )
        FreeRTOS_printf( ( "xNetworkInterfaceOutput failed. Link down?\n" ) 
#else 
        ;   /* No action */
#endif
}
```

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
AniruddhaKanhere: I am using the second style in +TCP since the first block comment in every file (the license) is of that format.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
AniruddhaKanhere: Ah, because memset has this signature:
'void *memset(void *str, int c, size_t n)'

gkwicker: Sorry I missed that, will resolve

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
AniruddhaKanhere: Done. I did not add the #else though. I hope its fine?

yuhui-zheng: I did not dig into how ```FreeRTOS_printf()``` is implemented, apologies for the ignorance there. 

Ideally, there should be some logger component, in which we could configure:
- which level of logging a module wants. (If +TCP wants verbose logging, then we could just configure it.) 
- where things log to. (could have multiple stream. UART, file, ...)
So the caller could avoid all the #if-#else blocks.

This implementation is much better than where we started, that I'm willing to get things going. Comment only, no change request.

htibosch:
Some comments about about logging and about the return value of `xNetworkInterfaceOutput()`:

When `ipconfigHAS_PRINTF` is non-zero, the macro `FreeRTOS_printf()` will produce output.
When `ipconfigHAS_DEBUG_PRINTF` is non-zero, the macro `FreeRTOS_debug_printf()` will produce output.

These macro's are only used within the FreeRTOS+TCP library and its drivers. A better name would have been `plus_tcp_printf()` and `plus_tcp_debug()`.

When I release a firmware, both macro's are disabled. This saves a lot of flash because the const logging strings are compiled.

Normally in the code, you don't have to check `ipconfigHAS_PRINTF` or `ipconfigHAS_DEBUG_PRINTF`.

In the following situation they are useful:

    #if( ipconfigHAS_DEBUG_PRINTF != 0 )
    {
        pcType = "UDP";
    }
    #endif /* ipconfigHAS_DEBUG_PRINTF != 0 */

The variable `pcType` is used for debugging the CRC-function.

@yuhui-zheng wrote:

- level of logging a module wants. (If +TCP wants verbose logging, then we could just configure it.)
- where things log to. (could have multiple stream. UART, file, ...)
  So the caller could avoid all the #if-#else blocks.

The #if-#else blocks rarely needed in the code. The two logging macro's are defined in `FreeRTOSIPConfig.h`.


Logging about `xNetworkInterfaceOutput()` is not really necessary because all changes in the Link Status are  already logged in the Network Interface.

yuhui-zheng:
@htibosch Thank you, that clarifies a lot. 


The very first comment about logging 
> xNetworkInterfaceOutput is implemented by third party, that I would check the return value to be sure the API contract is met.

was motivated by the understanding that it's usually vendors/third parties implementing +TCP porting layer including ```xNetworkInterfaceOutput```.  If there's a suite of test to verify porting (i.e. verify APIs input/output and behavior are compliant with spec), the confidence level will be much higher even without checking return values. The hope here is to fail fast, if there's any issue. 

Your concern about memory is valid, and to your other comment
> Logging about xNetworkInterfaceOutput() is not really necessary because all changes in the Link Status are already logged in the Network Interface.

This seems to imply we have a porting test suite that I'm not aware of. Or maybe my understanding about "it's usually vendors/third parties implementing +TCP porting layer" is incorrect. (Correct me here.)

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -86,8 +86,12 @@ static TickType_t xLastGratuitousARPTime = ( TickType_t ) 0;
...
 
 /* Part of the Ethernet and ARP headers are always constant when sending an IPv4
-ARP packet.  This array defines the constant parts, allowing this part of the
-packet to be filled in using a simple memcpy() instead of individual writes. */
+ * ARP packet.  This array defines the constant parts, allowing this part of the
+ * packet to be filled in using a simple memcpy() instead of individual writes. */
+
+/* Rule has been relaxed since this is a configuration parameter. And this location provides
+ * more readability. */
+/* coverity[misra_c_2012_rule_8_9_violation] */
```
gkwicker, line 94: I see no reason to suppress rule 8.9 here.  This is defined as `static` so it's not visible outside this module and it's referred to [only in a single location.](https://github.com/FreeRTOS/FreeRTOS/blob/master/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c#L627)  Readability will be improved by placing this definition in the block scope of `vARPGenerateRequestPacket()`, allowing you to delete lines 92-94 and eliminate this rule suppression.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -219,17 +223,17 @@ uint8_t ucMinAgeFound = 0U;
 
 	#if( ipconfigARP_STORES_REMOTE_ADDRESSES == 0 )
 		/* Only process the IP address if it is on the local network.
-		Unless: when '*ipLOCAL_IP_ADDRESS_POINTER' equals zero, the IP-address
-		and netmask are still unknown. */
+		 * Unless: when '*ipLOCAL_IP_ADDRESS_POINTER' equals zero, the IP-address
+		 * and netmask are still unknown. */
 		if( ( ( ulIPAddress & xNetworkAddressing.ulNetMask ) == ( ( *ipLOCAL_IP_ADDRESS_POINTER ) & xNetworkAddressing.ulNetMask ) ) ||
-			( *ipLOCAL_IP_ADDRESS_POINTER == 0ul ) )
+			( *ipLOCAL_IP_ADDRESS_POINTER == 0uL ) )
```
gkwicker, line 229: Why `uL` vs `UL`?  Let's be consistent and use the same convention everywhere, I will suggest `UL`.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
```diff
@@ -256,52 +260,56 @@ uint8_t ucMinAgeFound = 0U;
...
+					xARPCache[ x ].ucValid = ( uint8_t ) pdTRUE;					
 					return;
 				}
 
 				/* Found an entry containing ulIPAddress, but the MAC address
-				doesn't match.  Might be an entry with ucValid=pdFALSE, waiting
-				for an ARP reply.  Still want to see if there is match with the
-				given MAC address.ucBytes.  If found, either of the two entries
-				must be cleared. */
+				 * doesn't match.  Might be an entry with ucValid=pdFALSE, waiting
```
gkwicker, line 272: These comments are inconsistent when representing equality; see the comment on line 254.  I'd suggest modifying this comment as `ucValid == pdFALSE`.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
AniruddhaKanhere: I believe pc-lint likes a small 'u'. But both lint and coverity like big 'L'. Thats why 'uL' instead of other combinations...

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
AniruddhaKanhere: done...

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
AniruddhaKanhere: done

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_ARP.c
AniruddhaKanhere: Sticking with 'UL'

yuhui-zheng:
Above should've been a squash merge, clicked too fast. My bad. Instead of further messing with commit history, I'm going to leave it this time.

htibosch:
> This seems to imply we have a porting test suite that I'm not aware of.
> Or maybe my understanding about "it's usually vendors/third parties
> implementing +TCP porting layer" is incorrect. (Correct me here.)

and

>> xNetworkInterfaceOutput is implemented by third party

Many +TCP porting layers were developed and maintained by me. Beside that, I checked and tested most other implementations.

As far as I know, there is no document that explains how to test the porting layer.
It would also be good to have a guide that describes how to make an efficient network interface. Although there are a few good examples.

Here is how a network interface can be tested:

1) The simplest test is to **ping** the device.

2) Use **ping flood** ( `ping -f` from Linux ). That command sends as many echo messages per second as possible.

3) **iperf3**: the FreeRTOS+TCP device becomes an iperf server. Packets can be sent in both ways simultaneously.

4) **aws_tests** : makes use of the porting layer to communicate with an echo server.

5) **aws_demo** : uses DNS and makes connection a connection to an MQTT broker.

Now i stop, because the title of this PR is "MISRA compliance", and not "about porting +TCP".

Thanks, Hein

PR #55, opened 2020-04-13
ATmega port demo with simulator

yuhui-zheng:
Description
-----------
Copy-pasting below from README.md included in this PR. 

All changes in this PR are contained in ```./FreeRTOS/Demo/AVR_ATMega_AtmelStudio_simulator```. I don't see an obvious way to break down this PR. 

-------- from README.md --------
### Introduction
This demo project is trying to verify FreeRTOS ATmegaxxxx port implementation in the absense of hardware. Atmel Studio simulator is used in this project to verify clock/timing and task stack implementation. The focus is to ensure register value correctness, instead of high level software behavior correctness, since the simulation runs relatively slow. 

The simulator does not yet provide UART console simulation. Thus to verify point of interests user needs to set break points. 

#### Setting up environment
Download Atmel Studio IDE. Once finish IDE installation, also consider installing FreeRTOS Viewer plugin. 

#### To debug
Double click on ```AVR_ATMega_AtmelStudio_Simulator.atsln```. Follow instructions in AVR MCU Simulator debugging. 

#### To update target and/or peripherals
In Solution Explorer window, right click on RTOSDemo project --> Re-Configure Atmel Start Project. 

### Reference
- [Atmel Studio IDE](https://www.microchip.com/mplab/avr-support/atmel-studio-7)
- [FreeRTOS Viewer plugin](http://atmel-studio-doc.s3-website-us-east-1.amazonaws.com/webhelp/GUID-03D128BB-5CF6-42CB-8787-CB9A863C8F09-en-US-1/index.html)
- [AVR MCU Simulator debugging](http://atmel-studio-doc.s3-website-us-east-1.amazonaws.com/webhelp/GUID-54E8AE06-C4C4-430C-B316-1C19714D122B-en-US-1/index.html?GUID-C73F1111-250E-4106-B5E5-85A512B75E8B)

-------- end --------

Test Steps
-----------
- Get a workspace with this repo, and kernel updated to https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/48 (if not merged, otherwise master of FreeRTOS-kernel). 
- (Optional) To test with 8-bit timer, make sure ```portUSE_TIMER0``` is defined in ```port macro.h```. Or the default is to use watchdog timer. 
- Follow "To debug" section above. Setup break points to verify tick and task stack. 
Related Issue
-----------
N/A. Simulator is running slow. So that this project is not intended for long running test. 


yuhui-zheng:
A lot of files in this PR are generated by Atmel Studio IDE, and not all of them are needed by the demo. (You are right, the example you linked ```nvmctrl_basic_example.c``` is not needed by demo.)

Originally I was thinking "maybe Atmel users will recognize things they are already familiar with if we don't change the project structure too much". But I agree that at least we can strip out some example files. 

A couple of things I'll do:
- Keep driver files, even though they are not referenced in demo. It's mostly for backwards compatibility and we have some precedent in our code base. e.g. https://github.com/FreeRTOS/FreeRTOS/tree/master/FreeRTOS/Demo/CORTEX_MPU_STM32L4_Discovery_GCC_IAR_Keil/ST_Code

- I'll try to remove things other than driver, especially examples. Will also play with START to reconfigure project to see if those files will be brought back. (If necessary, I'll put something in .gitignore.)

Thank you for the review.

RichardBarry:
The project should provide everything necessary to build the demo, so users get everything they need from one place, but nothing more (i.e. not files that are not actually used by the demo).  

Originally third party driver source files were kept in a central location so there was only one copy used by all demos that needed that particular driver, but that was not maintainable, so now the driver source files that are actually used by the demo should be provided in the project's directory tree.  This also avoids changes in third party files breaking the builds, which could happen, if they were pulled in externally.

I think Atmel Start just provides the files your Start configuration requests, so as we have the Atmel Start meta data included in this PR we should ensure it can be used by everybody (i.e. it is not Atmel Account specific, meaning they were unusable by other users).

Preference is for the demo directory to be removed, with the FreeRTOSConfig.h config file it contains being moved either into the root where main.c is, or into the config directory, and for the regtest.c/h to also be moved to the same directory as the main.c file.

yuhui-zheng:
@RichardBarry  Noted, thank you.

For me:
- Will move ```FreeRTOSConfig.h``` and ```regtest.{c, h}``` to a proper location. (```regtest.{c,h}``` are very different from what's in ATMega323, that probably needs some scrutiny. Will put down some reasoning in comment next.)

- Two pieces of hw just arrived, I'll give those a try. If things work well, I'll rename the project for both emulator and hardware support.

yuhui-zheng:
I'm closing this PR, since https://github.com/FreeRTOS/FreeRTOS/pull/70. 

All the feedback above will be addressed in the new PR. The new PR runs on both HW and simulator.

PR #54, opened 2020-04-13
[Draft] CBMC proofs for +TCP

AniruddhaKanhere:
Draft PR for addition of +TCP CBMC proofs to this repo

Description
-----------
This PR aims to bring all the +TCP CBMC proofs in this repository. The end goal is to execute the proofs with every new PR to check whether the PR introduces any breaking change.
This is a PR that will involve quite a bit of editing and changes to allow CBMC to execute with the different directory structure of this repo (when comparing with amazon-freertos). I will make changes to it and people can review it and suggest changes while I am doing so.
Once everything is finalised, I will create a 'squashed' PR.

Test Steps
-----------
Once the files are edited to work with this repository,
- Run `python3 prepare.py` in the /FreeRTOS/Test/CBMC/proofs directory
- Commit the changes caused be prepare.py
- Run 'make' in each proof subdirectory

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #56, opened 2020-04-15
PR for addition of +TCP CBMC proofs to FreeRTOS/FreeRTOS

AniruddhaKanhere:
PR for addition of +TCP CBMC proofs to this repo

Description
-----------
- This PR aims to bring all the +TCP CBMC proofs in this repository. The end goal is to execute the proofs with every new PR to check whether the PR introduces any breaking change.
This is a PR that will involve quite a bit of editing and changes to allow CBMC to execute with the different directory structure of this repo (when comparing with amazon-freertos). I will make changes to it and people can review it and suggest changes while I am doing so.

- This PR doesn't add proofs for DNSgetHostByName, DNSgetHostByName_a, DNSHandlePacket, ReadNameField, SkipNameField, CheckOptionsInner and CheckOptionsOuter. These depend on configuration file and are not working yet. Once working, they'll be added incrementally.

Squashed from https://github.com/FreeRTOS/FreeRTOS/pull/54

Test Steps
-----------
Once the files are edited to work with this repository,
- Run `python3 prepare.py` in the /FreeRTOS/Test/CBMC/proofs directory
- Commit the changes caused be prepare.py
- Run 'make' in each proof subdirectory


FreeRTOS/Test/CBMC/patches/FreeRTOSConfig.h
```diff
@@ -134,6 +134,106 @@ extern void vAssertCalled( const char * pcFile,
...
+#define configINCLUDE_DEMO_DEBUG_STATS       1
+
+/* The size of the global output buffer that is available for use when there
+ * are multiple command interpreters running at once (for example, one on a UART
+ * and one on TCP/IP).  This is done to prevent an output buffer being defined by
+ * each implementation - which would waste RAM.  In this case, there is only one
+ * command interpreter running, and it has its own local output buffer, so the
+ * global buffer is just set to be one byte long as it is not used and should not
+ * take up unnecessary RAM. */
+#define configCOMMAND_INT_MAX_OUTPUT_SIZE    1
```
yuhui-zheng, line 160: This macro is referenced in ```FreeRTOS-Plus-CLI//FreeRTOS_CLI.c```, and we are not using +CLI as dependency here. Could remove.

FreeRTOS/Test/CBMC/patches/FreeRTOSConfig.h
```diff
@@ -134,6 +134,106 @@ extern void vAssertCalled( const char * pcFile,
...
+/* This demo creates a virtual network connection by accessing the raw Ethernet
+ * or WiFi data to and from a real network connection.  Many computers have more
+ * than one real network port, and configNETWORK_INTERFACE_TO_USE is used to tell
+ * the demo which real port should be used to create the virtual port.  The ports
+ * available are displayed on the console when the application is executed.  For
+ * example, on my development laptop setting configNETWORK_INTERFACE_TO_USE to 4
+ * results in the wired network being used, while setting
+ * configNETWORK_INTERFACE_TO_USE to 2 results in the wireless network being
+ * used. */
+#define configNETWORK_INTERFACE_TO_USE       ( 0L )
```
yuhui-zheng, line 175: If this macro is needed in proofs, we may need to understand why this is needed. I believe +TCP proofs intend to verify the logic of TCP/IP stack. Not quite about whether driver code is implemented correctly. So regardless of which physical/link layer we are using, there should be no impact on TCP layer. 

If removing this macro produces the same results, I'd suggest removing this line.

FreeRTOS/Test/CBMC/patches/FreeRTOSConfig.h
```diff
@@ -134,6 +134,106 @@ extern void vAssertCalled( const char * pcFile,
...
+
+/* The address of an echo server that will be used by the two demo echo client
+ * tasks:
+ * http://www.freertos.org/FreeRTOS-Plus/FreeRTOS_Plus_TCP/TCP_Echo_Clients.html,
+ * http://www.freertos.org/FreeRTOS-Plus/FreeRTOS_Plus_TCP/UDP_Echo_Clients.html. */
+#define configECHO_SERVER_ADDR0              192
```
yuhui-zheng, line 181: Between line 181 - 226. I don't think these are all necessary in TCP/IP stack proofs. And for some, I don't think fixing the addresses are clean. After all we are trying to verify TCP/IP stack logic. 

(e.g. making sure ARP works with any MAC address is one of the many goals... This seems obvious and no brain, but we want to demonstrate such with [formal methods](https://en.wikipedia.org/wiki/Formal_methods).)

In code during test setup, it'll look somewhat like for example  -- 
```
mac_addr0 = nondet_int8_t()
``` 
or 
```
__CPROVER_assume( mac_addr0 > 0 )
__CPROVER_assume( mac_addr0 <255 )
```
(if to leave out say 0.0.0.0 and 255.255.255.255, or any specific range.)

Proposing to take two steps:
- Remove unnecessary macros between 181-226. 
- After this PR gets merged, relax constraint on the proof inputs.

FreeRTOS/Test/CBMC/patches/FreeRTOSConfig.h
```diff
@@ -134,6 +134,106 @@ extern void vAssertCalled( const char * pcFile,
...
+
+/* Default netmask configuration.  Used in ipconfigUSE_DHCP is set to 0, or
+ * ipconfigUSE_DHCP is set to 1 but a DNS server cannot be contacted. */
+#define configNET_MASK0                      255
+#define configNET_MASK1                      255
+#define configNET_MASK2                      255
+#define configNET_MASK3                      0
+
+/* The UDP port to which print messages are sent. */
+#define configPRINT_PORT                     ( 15000 )
```
yuhui-zheng, line 229: This may not be essential to TCP proofs. Could leave out.

FreeRTOS/Test/CBMC/patches/FreeRTOSConfig.h
```diff
@@ -134,6 +134,106 @@ extern void vAssertCalled( const char * pcFile,
...
+
+#define configPROFILING                      ( 0 )
```
yuhui-zheng, line 231: I don't think this is necessary. Did not find reference.

FreeRTOS/Test/CBMC/patches/FreeRTOSIPConfig.h
```diff
@@ -0,0 +1,307 @@
...
+ * http://www.freertos.org/Stacks-and-stack-overflow-checking.html. */
+#define ipconfigIP_TASK_STACK_SIZE_WORDS           ( configMINIMAL_STACK_SIZE * 5 )
+
+/* ipconfigRAND32() is called by the IP stack to generate random numbers for
+ * things such as a DHCP transaction number or initial sequence number.  Random
+ * number generation is performed via this macro to allow applications to use their
+ * own random number generation method.  For example, it might be possible to
+ * generate a random number by sampling noise on an analogue input. */
+extern uint32_t ulRand();
+#define ipconfigRAND32()    ulRand()
```
yuhui-zheng, line 103: In ```FreeRTOSConfig.h``` we have ```configRAND32```. And here we have ```ipconfigRAND32```. You know the code better than I do, if possible could remove one.

FreeRTOS/Test/CBMC/patches/FreeRTOSIPConfig.h
```diff
@@ -0,0 +1,307 @@
...
+#define ipconfigTCP_KEEP_ALIVE_INTERVAL          ( 20 ) /* Seconds. */
+
+/* The socket semaphore is used to unblock the MQTT task. */
+#define ipconfigSOCKET_HAS_USER_SEMAPHORE        ( 0 )
+
+#define ipconfigSOCKET_HAS_USER_WAKE_CALLBACK    ( 1 )
+#define ipconfigUSE_CALLBACKS                    ( 0 )
+
+
+#define portINLINE                               __inline
```
yuhui-zheng, line 301: This shouldn't be here. The "inline" keyword is compiler dependent, and should be defined in kernel port not ```FreeRTOSIPConfig.h```

FreeRTOS/Test/CBMC/patches/FreeRTOSIPConfig.h
```diff
@@ -0,0 +1,307 @@
...
+
+void vApplicationMQTTGetKeys( const char ** ppcRootCA,
```
yuhui-zheng, line 303: Please leave out library related.

FreeRTOS/Test/CBMC/patches/FreeRTOSIPConfig.h
```diff
@@ -0,0 +1,307 @@
...
+*****************************************************************************/
+
+#ifndef FREERTOS_IP_CONFIG_H
+#define FREERTOS_IP_CONFIG_H
+
+/* Set to 1 to print out debug messages.  If ipconfigHAS_DEBUG_PRINTF is set to
+ * 1 then FreeRTOS_debug_printf should be defined to the function used to print
+ * out the debugging messages. */
+#define ipconfigHAS_DEBUG_PRINTF    0
+#if ( ipconfigHAS_DEBUG_PRINTF == 1 )
```
yuhui-zheng, line 41: For #if-#endif at line 41, 50 — if we don't need logging in proofs, please remove.

AniruddhaKanhere:
\bot run checks

FreeRTOS/Test/CBMC/README.md
```diff
@@ -1,133 +1,133 @@
...
-
-- `proofs` contains the proofs run against each pull request
-- `patches` contains a set of patches that get applied to the codebase prior to
-  running the proofs
-- `include` and `windows` contain header files used by the proofs.
+CBMC Proof Infrastructure
+=========================
+
+This directory contains automated proofs of the memory safety of various parts
+of the amazon:FreeRTOS codebase. A continuous integration system validates every
```
lundinc2, line 5: FreeRTOS

FreeRTOS/Test/CBMC/README.md
```diff
@@ -1,133 +1,133 @@
...
+of proof.
+
+
+CMake-based build
+-----------------
+
+Follow the CBMC installation instructions below.
+
+Suppose that the amazon-freertos source tree is located at
+`~/src/amazon-freertos` and you wish to build the proofs into
```
lundinc2, line 31: Need to change this to FreeRTOS paths and commands

FreeRTOS/Test/CBMC/README.md
```diff
@@ -1,133 +1,133 @@
...
+```
+If you are on a Windows machine but want to generate Linux Makefiles (or vice
+versa), you can pass the `--system linux` or `--system windows` options to those
+programs.
+
+
+### Running the proofs
+
+Each of the leaf directories under `proofs` is a proof of the memory
+safety of a single entry point in a:FR. The scripts that you ran in the
```
lundinc2, line 118: FreeRTOS

FreeRTOS/Test/CBMC/include/aws_freertos_ip_verification_access_ip_define.h
```diff
@@ -1,3 +1,3 @@
-eFrameProcessingResult_t publicProcessIPPacket( IPPacket_t * const pxIPPacket, NetworkBufferDescriptor_t * const pxNetworkBuffer ) {
-	prvProcessIPPacket(pxIPPacket, pxNetworkBuffer);
-}
```
lundinc2, line 3: Is this PR modifying line endings? I don't see any change.

AniruddhaKanhere: Yes. It is. Because the patches depend on the line endings. Therefore I had to modify the line ending to make the patches work.

yuhui-zheng:
Regarding to the previous commit to change file privilege from o755 -> o644. Please double check it is intended for files end with ```.py``` as well. (Folks in CI team may have previous experience about related. )

(I deleted duplicated "/bot run checks" for readability.)

FreeRTOS/Test/CBMC/include/aws_freertos_ip_verification_access_ip_define.h
AniruddhaKanhere: I went ahead and made all line endings conform to the previous standards. This removed ~100 files from the changed list.

AniruddhaKanhere:
> Regarding to the previous commit to change file privilege from o755 -> o644. Please double check it is intended for files end with `.py` as...

I observed that some python files have the execution bit set (https://github.com/FreeRTOS/FreeRTOS/pull/56/files#diff-51aff79ac3dbf526db62ba57ae17488c) and some don't (https://github.com/FreeRTOS/FreeRTOS/pull/56/files#diff-058b254c76a18c4674e9a471129dd30b). There was not a common standard. To improve on that, I set the execution bit on all python files in the CBMC directory.

FreeRTOS/Test/CBMC/README.md
AniruddhaKanhere: Done. Thanks for spotting that :)

FreeRTOS/Test/CBMC/README.md
AniruddhaKanhere: done

FreeRTOS/Test/CBMC/README.md
AniruddhaKanhere: Done

yuhui-zheng:
Since we have multiple people looking at this PR, posting status here. 

We are looking into why the number of checks doesn't sum up. (Before adding TCP proofs, it was ~46. After adding TCP proofs, I was expecting it to be much larger.) After that I'll just glance over the results in CI (folks may have already done this for a couple of time), and we should be good to go.

yuhui-zheng:
Another update on the failing proof -- ```ParseDNSReply```. 

cbmc.txt suggests we have a loop unwinding assertion failure at the 3rd loop in prvParseDNSReply. 

- Error message: [prvParseDNSReply.unwind.2] line 1078 unwinding assertion loop 2: FAILURE. 
- Source code: https://github.com/aws/amazon-freertos/blob/master/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c#L1078) 

We suspect this is related to how ```pucUDPPayloadBuffer``` is initialized. As an experiment:
Changing:
```
    uint8_t *pucUDPPayloadBuffer = buffer + buffer_offset;
```
to 
```
    DNSMessage_t xDnsMessage;
    __CPROVER_assume( xDnsMessage.usAnswers < 10); // bound it to some value. 10 for test purpose.
    uint8_t *pucUDPPayloadBuffer = (uint8_t)&xDnsMessage;
```
This resolves the unwinding issue. (The issue we are talking about here is not bug in source code. It's purely how the test is written.)

Since this round is to get things going, taking this proof out. ```ParseDNSReply``` and the other seven proofs shall be included in the subsequent PRs.

AniruddhaKanhere:
I have removed ParseDNSReply from the commit as it was failing. Also made changes to remove the same from AFR repo. Except this check, all others are present and working.

Subsequent PRs will add ParseDNSReply back to the checks.

PR #59, opened 2020-04-22
[CBMC]QueueGenericCreateStatic - Some assumptions did not match asserts in kernel source

dachalco:
Description
-----------
See the related asserts in [queue.c:320](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/master/queue.c#L320)

Logically equal, but easier to visually verify with the code.

Test Steps
-----------
Run cbmc tests in FreeRTOS/Test/CBMC/proofs/Queue/QueueGenericCreateStatic

Related Issue
-----------

### DIFF
--- a/FreeRTOS/Test/CBMC/proofs/Queue/QueueGenericCreateStatic/QueueGenericCreateStatic_harness.c
+++ b/FreeRTOS/Test/CBMC/proofs/Queue/QueueGenericCreateStatic/QueueGenericCreateStatic_harness.c
@@ -52,8 +52,10 @@ void harness(){
     __CPROVER_assume(uxQueueLength > ( UBaseType_t ) 0);
 
     // QueueGenericCreateStatic asserts the following equivalence
-    __CPROVER_assume( ( pucQueueStorage && uxItemSize ) ||
-                      ( !pucQueueStorage && !uxItemSize ) );
+    /* A queue storage area should be provided if the item size is not 0, and
+    should not be provided if the item size is 0. */
+    __CPROVER_assume( !( ( pucQueueStorage != NULL ) && ( uxItemSize == 0 ) ) );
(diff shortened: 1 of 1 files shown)


FreeRTOS/Test/CBMC/proofs/Queue/QueueGenericCreateStatic/QueueGenericCreateStatic_harness.c
```diff
@@ -52,8 +52,10 @@ void harness(){
     __CPROVER_assume(uxQueueLength > ( UBaseType_t ) 0);
 
     // QueueGenericCreateStatic asserts the following equivalence
-    __CPROVER_assume( ( pucQueueStorage && uxItemSize ) ||
-                      ( !pucQueueStorage && !uxItemSize ) );
+    /* A queue storage area should be provided if the item size is not 0, and
```
cobusve, line 55: This may help : https://en.wikipedia.org/wiki/De_Morgan%27s_laws

dachalco: oops. Yes I did use demorgans to evaluate a few transformations but clearly not enough. With more transformations, they are logically equal. My mistake, preemptively assumed inequality after getting about halfway through the necessary transformations.

Though, I'd argue it's still better to change them because it requires reader to do several boolean transformations to verify it matches the code

dachalco:
Closing this. Silly mistake on my part drawing conclusion from a glance...thanks for checking!

My mistake highlights a different issue. There's confusing code/comments in [this kernel code](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/master/queue.c#L320). It would make way more sense to have
```c
configASSERT( ( pucQueueStorage && uxItemSize ) ||
             ( !pucQueueStorage && !uxItemSize ) );
```
in the kernel source, especially following the code comment linked above which grammatically expresses the logic in the simpler form.

PR #64, opened 2020-04-28
Move CBMC proofs to FreeRTOS+ directory

AniruddhaKanhere:
PR to move +TCP CBMC proofs to correct directory (i.e. FreeRTOS-Plus).

Description
-----------
This PR moves the +TCP proof which exist in the FreeRTOS directory to FreeRTOS-Plus/Test directory.
The other supporting files are duplicated with this PR. This will remain so till a common place for the CBMC-setup files is decided.

Added changes to Makefiles to reflect the change in the directory structure.

Also, added the proof for ParseDNSReply which was failing earlier.


FreeRTOS-Plus/Test/CBMC/include/queue_init.h
```diff
@@ -0,0 +1,141 @@
...
+
+	QueueHandle_t xQueue =
+		xQueueGenericCreate(uxQueueLength, uxItemSize, ucQueueType);
+	if(xQueue){
+		xQueue->cTxLock = nondet_int8_t();
+		xQueue->cRxLock = nondet_int8_t();
+		xQueue->uxMessagesWaiting = nondet_UBaseType_t();
+		/* This is an invariant checked with a couple of asserts in the code base.
+		   If it is false from the beginning, the CBMC proofs are not able to succeed*/
+		__CPROVER_assume(xQueue->uxMessagesWaiting < xQueue->uxLength);
```
yuhui-zheng, line 80: Just a headsup. I'm cleaning up ```queue_init.h``` in an other PR. In your PR we are duplicating queue/task stubs and initializations, I understand some of them are necessary. But we may get queue/task common structures out of sync between ```./FreeRTOS/Test``` and ```./FreeRTOS-Plus/Test```.


I'm commenting on this specific line, since this is exactly what I'm trying to clean up in another PR. ```xQueue->uxMessagesWaiting``` is bounded at line 77. And here we are assuming with a comparison to an unbounded member ```xQueue->uxLength```...

FreeRTOS-Plus/Test/CBMC/patches/FreeRTOSConfig.h
```diff
@@ -0,0 +1,237 @@
...
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * http://aws.amazon.com/freertos
+ * http://www.FreeRTOS.org
+ */
+
+#ifndef FREERTOS_CONFIG_H
```
yuhui-zheng, line 26: I'm thinking the content of this file should be the same in ```./FreeRTOS/Test``` and ```./FreeRTOS-Plus/Test```. This stays as a comment, no specific action item on you. We'll need some automated checks to ensure this later. 


Also, is this PR migrating from the same repo? Or migrating from aws/freertos-amazon again? The reason for asking is we have lots of addresses hard coded in this file....

FreeRTOS-Plus/Test/CBMC/patches/FreeRTOSIPConfig.h
```diff
@@ -0,0 +1,307 @@
...
+
+/*****************************************************************************
+*
+* See the following URL for configuration information.
+* http://www.freertos.org/FreeRTOS-Plus/FreeRTOS_Plus_TCP/TCP_IP_Configuration.html
+*
+*****************************************************************************/
+
+#ifndef FREERTOS_IP_CONFIG_H
+#define FREERTOS_IP_CONFIG_H
```
yuhui-zheng, line 35: Also a comment without action item. This file is not needed for kernel proofs. Can put whichever makes sense to TCP proofs.

FreeRTOS-Plus/Test/CBMC/patches/FreeRTOSConfig.h
AniruddhaKanhere: This is migrating from the same repo. I have just moved the files from FreeRTOS to FreeRTOS/Plus (with some changes to allow the code to function as expected).

FreeRTOS-Plus/Test/CBMC/include/queue_init.h
AniruddhaKanhere: It seems I don't need queue_init.h at all. I removed it altogether.

PR #71, opened 2020-05-22
Test: Add Linux Networking support with demo application

alfred2g:
Add Linux Networking support with demo application

Description
-----------
1. Add networking support for FreeRTOS with libpcap
2. Add a demo application: tcp echo client
3. Run uncrustify on some files

Test Steps
-----------
Run Demo application by running
1. Modify the configuration files (FreeRTOSConfig.h and FreeRTOSIPConfig.h) and add IP's Gateways, echo server address etc..
Enable the echo ip client from main 
``` C
 #define mainCREATE_TCP_ECHO_TASKS_SINGLE    1
```
2. run an echo server on a different machine
``` 
sudo nc -l 7
```
3.
```
$ scons
$ sudo ./build/posix_demo
```
4. You are supposed to see some echo strings printed on the remote server running the echo server

5. There is knows issue, the echo server can't be on the same machine, it can be from a VM to a host though. For some reason, when it is on the same machine, the host is not sending ARP replies back to the demo application ip stack, which renders the demo application unable to connect to the echo server

Related Issue
-----------


FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
```diff
@@ -60,33 +60,34 @@ a constant. */
```
yuhui-zheng, line 61: Are the changes to ```FreeRTOS_IP.c``` necessary? It seems we are formatting this file.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/portable/NetworkInterface/WinPCap/NetworkInterface.c
```diff
@@ -44,20 +44,20 @@ access functions. */
```
yuhui-zheng, line 44: ```WinPCap/NetworkInterface.c``` does not seem to relate to Linux Simulator. I'd recommend having another PR to avoid confusion.

Issue #87, opened 2020-06-07
ICMP checksum offoloading [Inquiry]


e-sr:
I am tring to port the tcpip stack to the msp432E from ti. The emac peripheral can perform header checksum offloading thus i am using `ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM` set to 1.
My driver implementation works fine in the case of  ip header checksum offloading. However it does not work in the case of ICMP, pings are discarded beacuse of checksum error. Looking at the function  `prvProcessICMPEchoRequest`  in `FreeRTOS/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c` i realized that the ICMP  checksum is calculated despite of the option   `ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM`.
Commenting out the checksum part in `prvProcessICMPEchoRequest` as follow 
```c
/* due to compiler warning "integer operation result is out of range" */
#if ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM
pxICMPHeader->usChecksum=0;
#else
usRequest = ( uint16_t ) ( ( uint16_t )ipICMP_ECHO_REQUEST << 8 );

if( pxICMPHeader->usChecksum >= FreeRTOS_htons( 0xFFFFu - usRequest ) )
{
	pxICMPHeader->usChecksum = ( uint16_t )
		( ( ( uint32_t ) pxICMPHeader->usChecksum ) +
			FreeRTOS_htons( usRequest + 1UL ) );
}
else
{
	pxICMPHeader->usChecksum = ( uint16_t )
		( ( ( uint32_t ) pxICMPHeader->usChecksum ) +
			FreeRTOS_htons( usRequest ) );
}
#endif
```
solve the problem. Is it possible  to add an ipconfig defines 
`ipconfigDRIVER_INCLUDED_TX_ICMP_CHECKSUM` to simplify porting?

RichardBarry:
Would ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM and the suggested ipconfigDRIVER_INCLUDED_TX_ICMP_CHECKSUM ever be set differently?

[...]

Issue #65, opened 2020-04-29
Feature request: FreeRTOSConfig_Template.h


pieterconradie:
Hi there!

I'm new to FreeRTOS and am creating a test project for an STM32L072 + GCC that does not have a demo project specifically for it.

Each demo project has it's own FreeRTOSConfig.h file, but their ages vary and the content does not seem to keep up with development (?) I have found this page that appears to be up to date with the latest configuration settings (but I'm not 100% sure):
[Kernel > Developer Docs > FreeRTOSConfig.h](https://www.freertos.org/a00110.html)

I want to recommend that a FreeRTOSConfig_Template.h be created and kept updated and in sync with all of the possible configuration settings. That way I will know that I have all of the bases covered.

Thanks for your consideration,
Pieter
https://piconomix.com

PR #92, opened 2020-06-17
Driver checksum support for ICMP ping reply

Spadi0:
Description
-----------

When using `ipconfigREPLY_TO_INCOMING_PINGS`, it does not currently support letting the driver calculate the ICMP checksum, even when `ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM` is set to 1. It fills up the ICMP checksum field with data, which messes with the hardware checksum calculation (at least, it did for me whilst using an STM32F429).

Test Steps
-----------

1. Set `ipconfigREPLY_TO_INCOMING_PINGS` to 1
2. Use a hardware checksum calculation for ICMP packets

Related Issue
-----------
I found an issue that was already open, created by someone else (#87).


FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
```diff
@@ -1844,6 +1844,9 @@ uint8_t ucProtocol;
 
 		usRequest = ( uint16_t ) ( ( uint16_t )ipICMP_ECHO_REQUEST << 8 );
```
AniruddhaKanhere, line 1845: Hello,
Can you move this statement (along with the comment on line #1843) inside the `#else` as well?
Because in this case, the compiler will generate a warning saying `the operation result is unused` if `( ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM != 1 )`.
Thanks

AniruddhaKanhere: Also, @htibosch, can you have a look at this as well?

htibosch: Hello [Sam](https://github.com/Spadi0), I can not replicate the problem you mention both on a STM32F407 DISCO and a STM32F746G DISCO board.
I don't have a F429, but I think that the EMAC peripheral is the same,
I did this test both with `ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM` enabled and disabled.

@[Aniruddha](https://github.com/AniruddhaKanhere): your patch is correct, but this is the responsibility of the NetworkInterface. You can see it [here](https://github.com/FreeRTOS/FreeRTOS/blob/master/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/portable/NetworkInterface/STM32Fxx/NetworkInterface.c) around line 657:
:
~~~c
    if( pxPacket->xICMPPacket.xIPHeader.ucProtocol == ( uint8_t ) ipPROTOCOL_ICMP )
    {
        pxPacket->xICMPPacket.xICMPHeader.usChecksum = ( uint16_t )0u;
    }
~~~
Many EMACS require that the value of the ICMP checksum field is set to 0x00.
Some EMACS don't do ICMP checksums and they depend on the code you disabled here above.

@Spadi0 Now what makes you think that the ICMP checksum on the wire is not correct?
If you saw that in WireShark, you might be mislead. Most PC's and laptops have checksum offloading activated, and therefore Wireshark doesn't get to see the original checksum.

While developing, I recommend to switch off checksum offloading in the laptop and let Wireshark verify all checksums.

If you want you can attach (a zipped) PCAP file to your post.

Spadi0: Hi @htibosch,

I do see what you mean about some EMACs requiring the checksum value. My bad. However, I am not sure what to change it to. I feel that a new config option would need to be added.

About me being sure the checksum was invalid: it initially came about when I attempted to ping the device using my computer, and got no reply. After looking at Wireshark, it was saying the checksum was invalid - I now realise the checksum was being offloaded (as it was set to 0x00). After implementing the change which was commited, it started working, and I thought that the EMAC was not bothering to calculate a checksum as one was already inserted.
As it would turn out, my reasoning was wrong, but my solution wasn't. After disabling checksum offloading, Wireshark now displays a checksum (which is invalid). I have attached the zipped PCAP file for you.

As for you not being able to replicate this issue, I am not sure what to say. I am using the STM32429I-EVAL1, in case it is relevent. It is worth noting that I did not use the included STM32F4 NetworkInterface (as I was intentionally not using the STM32 HAL for this project, and thought it would be a good learning experience), and wrote my own. It functions fine (as far as I can tell, at least), but there may be some configuration option that the HAL uses which I don't in my own driver.

I feel it is still worth exploring this issue though, because like you said, many EMACs require the ICMP checksum field to be set to 0x00, and while it can be implemented in the driver (which could be done fairly easily by using the already provided NetworkInterface files as reference), this doesn't seem to be documented anywhere. And even if it was, it would be easier for an end user to add a configuration option than to implement it themselves.

I would be happy to do some more work on this, just require the guidance and direction.

Many thanks,
Sam

htibosch: > I do see what you mean about some EMACs requiring the checksum
> value. My bad. However, I am not sure what to change it to.
> I feel that a new config option would need to be added.

I don't think so, your NetworkInterface.c can solve this, as in the example that I showed.

I also have a NetworkInterface for STMFxx that doesn't make use of HAL but the so-called standard-peripheral driver in stead. Is that what you did as well?

Most users just grab a NetworkInterface and they expect it to work without a minimum of configuration.  Remember that `FreeRTOSIPConfig.h` must be maintained by the users.

I found several types of behaviour:

IP, TCP and UDP :
- checksum offloading work with any preparation.

ICMP :
- some EMAC works without any preparation
- most EMAC's ( STM32Fx ) expect us to clear the checksum-field
- some EMAC's can not do ICMP

When you look at the existing drivers, they all make sure that the ICMP checksum will be set, without depending on an extra macro.

You can look at the STM32Fxx driver to see how it was solved for the STM32F platform.

But if you want, you can send or share your driver and I will look at it.

e-sr: Is not the main  point of **checksum offloading** offload the mcu  and let it the hardware  calculate checksum instead? 
The example proposed is more a workaround than a solution. The target calculate then clear and after recalculate the checksum in hardware. 

However  i am not as good to code in a performat way :-( so is performance not an issue in my case...

> Most users just grab a NetworkInterface and they expect it to work without a minimum of configuration. Remember that FreeRTOSIPConfig.h must be maintained by the users.

I am new to freeRTOS and creating `FreeRTOSIPConfig.h` was not easy. The main concern is however  a missing template file as stated in #65. I don' think the difficulty will be some more macro if well documented.

AniruddhaKanhere:
@Spadi0, can you please recreate this PR in the new FreeRTOS-Plus-TCP repository [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP).
The FreeRTOS+TCP source code does not reside in FreeRTOS/FreeRTOS (this) repository anymore.

Apologies for the inconvenience.

AniruddhaKanhere:
Hello @Spadi0, it seems that @htibosch has suggested a fix for the problem being pointed out by you.
If the suggested solution doesn't work, feel free to re-open this PR or a new PR/Issue in the [FreeRTOS+TCP](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP) repository.

@Spadi0, Thanks again for pointing out the issue and taking the time to suggest and discuss a potential fix. I shall close this PR now.

PR #98, opened 2020-06-19
Tools: Add precommit git sccript to tools and uncrustify

alfred2g:
Add precommit git sccript to tools and uncrustify

Description
-----------
Add precommit git sccript to tools and uncrustify to be able to add style checks into Jenkins CI


tools/git/hooks/src/pre_commit.py
```diff
@@ -121,6 +121,8 @@ def is_checked_file_pattern(file_name):
 
 def is_ignored_file_pattern(file_name):
     ignored_patterns = [
+        r"FreeRTOS-Plus",
```
aggarw13, line 124: Why is the entire directory of `FreeRTOS-Plus ` ignored? Shouldn't it be the kernel and the FreeRTOS+TCP directories that are ignored?

PR #104, opened 2020-06-23
Add checks in FreeRTOS_Socket.c

AniruddhaKanhere:
Description
-----------
This PR adds changes to the FreeRTOS_Sockets.c to form a base for the Full-TCP test suite migration to FreeRTOS/FreeRTOS repository. The source is modified to add some checks in appropriate places such as `NULL` checking before dereferencing a pointer, defensive coding by assigning an error code to the return variable etc.


FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_Sockets.c
```diff
@@ -1397,6 +1397,12 @@ FreeRTOS_Socket_t *pxSocket;
 	( void ) uxOptionLength;
 
 	configASSERT( xSocket != NULL );
+	
+	if (pxSocket == FREERTOS_INVALID_SOCKET)
+	{
+		xReturn = FREERTOS_EINVAL;
+		return xReturn;
+	}
```
htibosch, line 1405: Firstly, mind the white space:
~~~c
- if (pxSocket == FREERTOS_INVALID_SOCKET)
+ if( pxSocket == FREERTOS_INVALID_SOCKET )
~~~
I think it is good to treat the a NULL pointer the same, returning a `-EINVAL`:
~~~c
if( ( pxSocket == NULL ) || ( pxSocket == FREERTOS_INVALID_SOCKET ) )
{
    xReturn = FREERTOS_EINVAL;
    return xReturn;
}
~~~

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_Sockets.c
```diff
@@ -2271,12 +2277,13 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t *pxSocket )
...
 	BaseType_t xTimed = pdFALSE;
-	BaseType_t xResult;
+	BaseType_t xResult = FREERTOS_EINVAL;
 	TimeOut_t xTimeOut;
 
 		( void ) xAddressLength;
 
-		xResult = prvTCPConnectStart( pxSocket, pxAddress );
+		if( pxAddress != NULL )
+			xResult = prvTCPConnectStart( pxSocket, pxAddress );
```
htibosch, line 2286: The assignment should be put into a compound block.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_Sockets.c
```diff
@@ -2758,8 +2766,9 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t *pxSocket )
 		/* Prevent compiler warnings about unused parameters.  The parameter
 		may be used in future versions. */
 		( void ) xFlags;
-
-		xByteCount = ( BaseType_t ) prvTCPSendCheck( pxSocket, uxDataLength );
+		
+		if (pvBuffer != NULL)
+			xByteCount = ( BaseType_t ) prvTCPSendCheck( pxSocket, uxDataLength );
```
htibosch, line 2771: Please make this a compound statement.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_Sockets.c
```diff
@@ -2758,8 +2766,9 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t *pxSocket )
 		/* Prevent compiler warnings about unused parameters.  The parameter
 		may be used in future versions. */
 		( void ) xFlags;
-
-		xByteCount = ( BaseType_t ) prvTCPSendCheck( pxSocket, uxDataLength );
+		
+		if (pvBuffer != NULL)
```
htibosch, line 2770:
~~~c
-if (pvBuffer != NULL)
+if( pvBuffer != NULL )
~~~

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_Sockets.c
```diff
@@ -2480,9 +2487,10 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t *pxSocket )
 
 		/* Check if the socket is valid, has type TCP and if it is bound to a
 		port. */
-		if( prvValidSocket( pxSocket, FREERTOS_IPPROTO_TCP, pdTRUE ) == pdFALSE )
+		if( ( prvValidSocket( pxSocket, FREERTOS_IPPROTO_TCP, pdTRUE ) == pdFALSE ) || 
+			( pvBuffer == NULL ) )
 		{
-			xByteCount = -pdFREERTOS_ERRNO_EINVAL;
+			xByteCount = FREERTOS_EINVAL;
```
htibosch: There is one situation in which it is valid to set `pvBuffer` to `NULL`: after the data have been read with the flag `FREERTOS_ZERO_COPY`.
Checking `prvValidSocket()` here should be enough.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_Sockets.c
AniruddhaKanhere: Oops about the white space. IDE adds them sometimes!
Also, thanks for the pointer about pxSocket being NULL. I made a commit to correct that.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_Sockets.c
AniruddhaKanhere: done

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_Sockets.c
AniruddhaKanhere: done, thanks!

htibosch: All good, thank you.

AniruddhaKanhere:
Thanks @yuhui-zheng for pointing that out. I'll remove it in a different PR. The Unit tests folder has many files which were copied there from FreeRTOS-Plus-TCP. I will remove them.

htibosch:
In `FreeRTOS_recv()`, when the zero-copy flag is set, it would be good to check the value of `pvBuffer`.
Around line 2494:
~~~c
	/* Check if the socket is valid, has type TCP and if it is bound to a
	port. */
	if( prvValidSocket( pxSocket, FREERTOS_IPPROTO_TCP, pdTRUE ) == pdFALSE )
	{
		xByteCount = -pdFREERTOS_ERRNO_EINVAL;
	}
+	else if( ( ( ( uint32_t ) xFlags & ( uint32_t ) FREERTOS_ZERO_COPY ) != 0U ) &&
+			 ( pvBuffer == NULL ) )
+	{
+		/* In zero-copy mode, pvBuffer is a pointer to a pointer ( not NULL ). */
+		xByteCount = -pdFREERTOS_ERRNO_EINVAL;
+	}
	else
	{
~~~
Beside that I think the PR is all good.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_Sockets.c
```diff
@@ -2271,12 +2275,15 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t *pxSocket )
...
 	TickType_t xRemainingTime;
 	BaseType_t xTimed = pdFALSE;
-	BaseType_t xResult;
+	BaseType_t xResult = -pdFREERTOS_ERRNO_EINVAL;
 	TimeOut_t xTimeOut;
 
 		( void ) xAddressLength;
 
-		xResult = prvTCPConnectStart( pxSocket, pxAddress );
+		if( pxAddress != NULL )
```
gkwicker, line 2283: Why are we validating `pxAddress` but not `pxSocket` or `xAddressLength`?

AniruddhaKanhere: `pxSocket` is validated by `prvTCPConnectStart` which checks fo the socket being `NULL` or `FREERTOS_INVALID_SOCKET`.

`xAddressLength` is not used at all in this implementation. See a couple of lines above this comment: `( void ) xAddressLength;` is used to ignore the parameter `xAddressLength`.

gkwicker: Then if `prvTCPConnectStart()` validates one parameter `pxSocket` why doesn't it validate both of them?  It is weird validating one parameter in each routine, why not just do the validations in one place?

AniruddhaKanhere: I agree with you on this front. `prvTCPConnectStart()` is used only in this particular function so it can be easily modified without any repercussions. I'll update that!
Thanks :)

AniruddhaKanhere: Done!

PR #118, opened 2020-07-02
MQTT Lightweight demo

leegeth:
MQTT Lightweight demo

Description
-----------
Demo for showing use of the MQTT light weight serializer API.
The Light weight serializer API lets user to serialize and
deserialize MQTT messages into a user provided buffer.
This API allows use of a statically allocated buffer.

The Example shown below uses this API to create MQTT messages and
send them over the connection established using FreeRTOS sockets.
The example is single threaded and uses statically allocated memory;
it uses QOS0 and therefore does not implement any retransmission
mechanism for Publish messages.


leegeth:
Opened a PR instead of raising a draft. Closing it and will be raising a draft PR

PR #122, opened 2020-07-04
Fix usGenerateChecksum on 8-bit platforms with odd-aligned buffers

andyjpb:
Fix usGenerateChecksum on 8-bit platforms with odd-aligned buffers

Description
-----------
 + Fix casting of pointers to integers when pointers aren't 32-bits wide.
 + Fix generation of checksum on odd-aligned buffers with non-zero initialisation vector.

Test Steps
-----------
On avr8 the stack complains about invalid checksums for arriving UDP and TCP packets until this patch is applied.

Surprisingly, the stack can receive the DHCP packets necessary to obtain an IP address but all other UDP and TCP packets seem to fail the UDP or TCP checksum. The IP header checksum is not affected.

Reproduce:
TCP:
...by trying to telnet to a port on the target
`$ telnet 172.25.11.35 51`

UDP:
...by using netcat to send packets to the target
`echo "Hello" | nc -u 172.25.11.35 51`

I was running this on an ATMega1284 with an Microchip ENC28J60 combined MAC and PHY.

I reproduced the issue by compiling the usGenerateChecksum function on Linux-buster-x86_64 and passing it captured packets that I had pre-padded to be aligned on different boundaries. I can supply this code if desired.


htibosch:
@andyjpb really fantastic that you have FreeRTOS running on an ATMega!
I will look at your PR later.
Thanks!

htibosch:
@andyjpb: I tested your PR on a Zynq 7000 platform, and it seems to work well. It should be tested on other platforms as well though.
I used these these defines in my `FreeRTOSIPConfig.h`:
~~~c
#define ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM		( 0 )
#define ipconfigDRIVER_INCLUDED_RX_IP_CHECKSUM		( 0 )
~~~
meaning that the IP-task will calculate the incoming and outgoing checksums.

It doesn't look like the ENC28J60 can do checksum offloading? I'm afraid that the 32-bit calculation has a lot of overhead for an ATMega128x.

htibosch:
One more question: didn't you have problems with these functions:
~~~c
NetworkBufferDescriptor_t *pxPacketBuffer_to_NetworkBuffer( const void *pvBuffer )
NetworkBufferDescriptor_t *pxUDPPayloadBuffer_to_NetworkBuffer( const void * pvBuffer )
~~~
They are used to go from a ( payload ) buffer to a network buffer. The data buffer stores a 32-bit pointer to the containing `NetworkBufferDescriptor_t`.

andyjpb:
Hi,

> One more question: didn't you have problems with these functions:
> 
> ```c
> NetworkBufferDescriptor_t *pxPacketBuffer_to_NetworkBuffer( const void *pvBuffer )

This one appears to only be compiled for zero copy drivers:

```
#if( ipconfigZERO_COPY_TX_DRIVER != 0 ) || ( ipconfigZERO_COPY_RX_DRIVER != 0 )
```

...which I haven't configured for my driver.


> ```
> NetworkBufferDescriptor_t *pxUDPPayloadBuffer_to_NetworkBuffer( const void * pvBuffer )

This one doesn't seem to be excluded from the build and doesn't generate a compiler warning.
I've not had any trouble with it yet but haven't yet done more than ping the target and use the DHCP client.

andyjpb:
Hi @htibosch,

> @andyjpb: I tested your PR on a Zynq 7000 platform, and it seems to work well. It should be tested on other platforms as well though.

Thanks for taking a look!


> It doesn't look like the ENC28J60 can do checksum offloading?

That's right as far as I know.


> I'm afraid that the 32-bit calculation has a lot of overhead for an ATMega128x.

There's perhaps a better algorithm but I wanted to get it working with the stock code first.
Would the way to provide a custom implementation be to switch on checksum offloading and then implement it in software in the driver?

I'm not too worried about performance at the moment: I'll profile things later when I have a better idea of how everything will perform together. It looks like I might be able to get some idea of the checksum overhead by implementing a simple UDP echo server and comparing timings for different packet sizes with the ICMP echo timings.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
```diff
@@ -2198,10 +2199,20 @@ size_t uxDataLengthBytes = uxByteCount;
...
 	/* The object pointer expression "pucNextData" of type "uint8_t const *" is cast to an integer type "unsigned int". */
-	ulAlignBits = ( ( ( uint32_t ) pucNextData ) & 0x03U ); /*lint !e9078 !e923*/	/* gives 0, 1, 2, or 3 */
+	uxAlignBits = ( ( intptr_t ) pucNextData ) & 0x03U; /*lint !e9078 !e923*/	/* gives 0, 1, 2, or 3 */
+
+	/*
+	 * If pucNextData is non-aligned then the checksum is starting at an
+	 * odd position and we need to make sure the usSum value now in xSum is
+	 * as if it had been "aligned" in the same way.
+	 */
+	if( ( uxAlignBits & 1UL) != 0U )
```
gkwicker, line 2209: For consistency, should use `!= 0UL` as on line 2215.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
```diff
@@ -2198,10 +2199,20 @@ size_t uxDataLengthBytes = uxByteCount;
...
+	{
+		xSum.u32 = ( ( xSum.u32 & 0xffU ) << 8 ) | ( ( xSum.u32 & 0xff00U ) >> 8 );
```
gkwicker, line 2211: Why is this using a 32-bit data type given that the checksum is only 16 bits?

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
```diff
@@ -2210,7 +2221,7 @@ size_t uxDataLengthBytes = uxByteCount;
 	}
 
 	/* If half-word (16-bit) aligned... */
-	if( ( ( ulAlignBits == 1U ) || ( ulAlignBits == 2U ) ) && ( uxDataLengthBytes >= 2U ) )
+	if( ( ( uxAlignBits == 1U ) || ( uxAlignBits == 2U ) ) && ( uxDataLengthBytes >= 2U ) )
```
gkwicker, line 2224: If it is aligned to a 16-bit boundary, wouldn't the test be?

```c
( uxAlignBits == 2UL ) || ( uxAlignBits == 0UL )
```

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
```diff
@@ -2289,7 +2300,7 @@ size_t uxDataLengthBytes = uxByteCount;
 	/* coverity[value_overwrite] */
 	xSum.u32 = ( uint32_t ) xSum.u16[ 0 ] + xSum.u16[ 1 ];
 
-	if( ( ulAlignBits & 1U ) != 0U )
+	if( ( uxAlignBits & 1U ) != 0U )
```
gkwicker, line 2303: The logic on lines 2304-2308 appears identical to the new logic added in lines 2210-2212.  Is this block still required?

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
```diff
@@ -2182,7 +2182,8 @@ aid though to optimise the calculations. */
 xUnion32 xSum2, xSum, xTerm;
 xUnionPtr xSource;
 xUnionPtr xLastSource;
-uint32_t ulAlignBits, ulCarry = 0UL;
+intptr_t uxAlignBits;
```
AniruddhaKanhere, line 2185: I noticed that you have used `intptr_t` here. While it is generally okay to do that, I would suggest exploring `BaseType_t`. Is `BaseType_t` defined for your platform (I know that it is there for all 16 and 32 bit platforms)?

andyjpb: Hi @AniruddhaKanhere 

Thanks for the feedback.

I have BaseType_t as an int8_t on my platform but pointers are 16bits.

I also have TickType_t which can be uint16_t or uint32_t depending on configUSE_16_BIT_TICKS but it didn't seem appropriate to use that for a pointer. It'll work, but a 32bit TickType_t will still generate a compiler warning.

What do you think?

andyjpb:
Thanks for the kind words and feedback @AniruddhaKanhere.

I'll look at @gkwicker's comments: some of them are quite detailed and relate to how the code was already structured so I'll have to take a more in depth look to work out if my changes have any implications for them. I tried to change as little as possible but perhaps there's an opportunity to tidy up and/or simplify?

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
andyjpb: I had the same question!
My research suggests that this is a common algorithm for calculating the checksum that gives some performance benefits on machines with word-sizes greater than 16 bits.

Part (C) "Parallel Summation" at https://tools.ietf.org/html/rfc1071#page-3 has some more detail.

The overall algorithm implementation is pre-existing work, it just seems to have an alignment bug on 8-bit platforms: if the additions are done in this parallel way then it's important to make sure that the correct parts of the packet end up in the correct part of the summation: as long as it's consistent it doesn't matter how they're arranged (Part (B) at the above link), but on 8-bit platforms the arrangement was inconsistent.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
andyjpb: My understanding of this part of the code is that it swaps the endianness of the lower half of the accumulator back to that of the machine.
The order of the bytes in the accumulator depends on both the endianness of the machine and the alignment of the packet. So if the packet is aligned on an odd boundary then the bytes in the two half words of the accumulator are in different orders to each other.
This seems to be an interaction between Parts (B) and (C) of https://tools.ietf.org/html/rfc1071#page-3

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
andyjpb: My understanding of this code is that the preceding code ensures that byte aligned buffers are half-word aligned and then this part of the code handles half-word aligned buffers that are originally that way or have become that way because of the preceding code.

The data length check is just a guard to ensure that there's at least a half word of data.

Buffers that are already word aligned (uxAlignBits == 0UL) don't need any special treatment before the parallel summing step.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
AniruddhaKanhere: Hello @andyjpb,
Thanks for looking into it. A follow-up question just out of curiosity - what is the width of `intptr_t` and did you use `sizeof` to get the size of pointers?

If we don't have a type which has an architecture/platform specific width, I suggest that we stick to the original datatype (`uint32_t`). Even on platforms with low address width, we can use `uint32_t` (it is an overkill but safe).

I am just concerned since `intptr_t` might not be supported by all compilers. And this might break something in the long run. `uint32_t` in my opinion is safer.

Thoughts?

AniruddhaKanhere:
@andyjpb, I understand your point of "changing as less as possible" - less chance of breaking something.

Had your changes been non-impacting to the logic, such as casting/comments etc., I would agree with you.
But in this case, your patch changes some of the logic which implores us to look more closely at the existing code and the logic of the original author.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
htibosch: > Why is this using a 32-bit data type given that the checksum is only 16 bits?

Yes, pure optimisation.  The algorithm tries to do as many bytes in a single operation as possible.
So it will look for a 16-bit alignment, and then to a 32-bit alignment.  Bytes will then be added four at s time ("Parallel Summation").

The tricky thing is: the alignment can only be calculated when a pointer is cast into a number. MISRA doesn't like that.  What I did is cast it to a `uin32_t`, @andyjpb proposes to cast it to a intptr_t in stead, which has the same size as a void *, while bit-tests are allowed. So I vote for this change.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
htibosch: @AniruddhaKanhere wrote:

>  And this might break something in the long run. uint32_t in my opinion is safer.

Isn't that what we had already? The function used to cast a pointer to a `uint32_t`, and store the 2 lower bits in `ulAlignBits`.
Now it will be cast to a variable which has the same size as a void pointer ( 16-bits on an ATmega ).

andyjpb: Hi @AniruddhaKanhere,

The width of intptr_t is guaranteed to be an integer large enough to hold the address of a pointer on the target platform. I didn't use sizeof.
intptr_t is part of stdint.h, introduced in C99, so will be available everywhere the other platform-independent *_t types (uint32_t, etc) are available.

I'm happy to change it back if you wish but it will generate a warning on all platforms with pointers that are not 32bits. It could also cause problems on platforms where pointers are larger than 32bits.

So, yes, you're right: intptr_t might not be available on all compilers but it should certainly be available everywhere that has uint32_t.

AniruddhaKanhere: Thanks @htibosch and @andyjpb for your feedback/answers and patience with my suggestions!

While I agree that `intptr_t` is a better solution than `uint32_t`, there is no precedence of using this datatype anywhere in the code. I just don't want to set a precedence for various new datatypes in the source.

If @htibosch and @gkwicker are okay with this, I have no further objection :)

htibosch:
I have integrated Andy's changes in my private source-tree already 3 weeks ago.  Every project that I test or run uses this library ( and I configured them to use the algorithm ).

Why does nothing change?
Because `malloc()` on 32-platforms return buffers that are at least 32-bit aligned. I think that the `malloc()` used by ATmega returns 8-bit aligned buffers.

The byte-swap takes only place if a buffer is less than 16-bit aligned:
~~~c
    if( ( uxAlignBits & 1UL ) != 0U )
    {
        xSum.u32 = ( ( xSum.u32 & 0xffU ) << 8 ) | ( ( xSum.u32 & 0xff00U ) >> 8 );
    }
~~~
I just ran a test: I called `usGenerateChecksum()` four times, with uxAlignBits varying from 0 to 3.
All outcomes were the correct, 4 times `0x224C` in my case.
This is what the test looks like:
~~~c
    int uxIndex;
    uint16_t usSums[ 4 ];
    for( uxIndex = 0; uxIndex < ARRAY_SIZE( usSums ); uxIndex++ )
    {
        /* Copy the buffer to an offset of 0, 1, 2, or 3. */
        memcpy( pucCopy + uxIndex, ucGoodDnsResponse, sizeof ucGoodDnsResponse );
        /* Calculate the checksum. */
        usSums[ uxIndex ] = usGenerateChecksum( 0x1234U, pucCopy + uxIndex, sizeof ucGoodDnsResponse );
    }
~~~
And just to be sure, I disabled the swap and the outcomes were incorrect: `0x224C 0x442A 0x224C 0x442A`.

I borrowed the testing packet from a TCP-unit testing module:

~~~c
uint8_t ucGoodDnsResponse[] =
{
    0xd7, 0x66, 0x81, 0x80, 0x00, 0x01, 0x00, 0x08, 0x00, 0x00, 0x00, 0x00, 0x0e, 0x61, 0x33, 0x37,
    0x62, 0x78, 0x76, 0x31, 0x63, 0x62, 0x64, 0x61, 0x33, 0x6a, 0x67, 0x03, 0x69, 0x6f, 0x74, 0x09,
    0x75, 0x73, 0x2d, 0x77, 0x65, 0x73, 0x74, 0x2d, 0x32, 0x09, 0x61, 0x6d, 0x61, 0x7a, 0x6f, 0x6e,
    0x61, 0x77, 0x73, 0x03, 0x63, 0x6f, 0x6d, 0x00, 0x00, 0x01, 0x00, 0x01, 0xc0, 0x0c, 0x00, 0x05,
    0x00, 0x01, 0x00, 0x00, 0x01, 0x2c, 0x00, 0x1e, 0x0c, 0x69, 0x6f, 0x74, 0x6d, 0x6f, 0x6f, 0x6e,
    0x72, 0x61, 0x6b, 0x65, 0x72, 0x09, 0x75, 0x73, 0x2d, 0x77, 0x65, 0x73, 0x74, 0x2d, 0x32, 0x04,
    0x70, 0x72, 0x6f, 0x64, 0xc0, 0x1b, 0xc0, 0x48, 0x00, 0x05, 0x00, 0x01, 0x00, 0x00, 0x00, 0xec,
    0x00, 0x45, 0x09, 0x64, 0x75, 0x61, 0x6c, 0x73, 0x74, 0x61, 0x63, 0x6b, 0x2a, 0x69, 0x6f, 0x74,
    0x6d, 0x6f, 0x6f, 0x6e, 0x72, 0x61, 0x6b, 0x65, 0x72, 0x2d, 0x75, 0x2d, 0x65, 0x6c, 0x62, 0x2d,
    0x31, 0x77, 0x38, 0x71, 0x6e, 0x77, 0x31, 0x33, 0x33, 0x36, 0x7a, 0x71, 0x2d, 0x31, 0x31, 0x38,
    0x36, 0x33, 0x34, 0x38, 0x30, 0x39, 0x32, 0x09, 0x75, 0x73, 0x2d, 0x77, 0x65, 0x73, 0x74, 0x2d,
    0x32, 0x03, 0x65, 0x6c, 0x62, 0xc0, 0x29, 0xc0, 0x72, 0x00, 0x01, 0x00, 0x01, 0x00, 0x00, 0x00,
    0x23, 0x00, 0x04, 0x22, 0xd3, 0x41, 0xdb, 0xc0, 0x72, 0x00, 0x01, 0x00, 0x01, 0x00, 0x00, 0x00,
    0x23, 0x00, 0x04, 0x22, 0xd3, 0x53, 0xe4, 0xc0, 0x72, 0x00, 0x01, 0x00, 0x01, 0x00, 0x00, 0x00,
    0x23, 0x00, 0x04, 0x22, 0xd3, 0xb6, 0x17, 0xc0, 0x72, 0x00, 0x01, 0x00, 0x01, 0x00, 0x00, 0x00,
    0x23, 0x00, 0x04, 0x22, 0xd6, 0xf5, 0xf0, 0xc0, 0x72, 0x00, 0x01, 0x00, 0x01, 0x00, 0x00, 0x00,
    0x23, 0x00, 0x04, 0x22, 0xd7, 0xe6, 0xa4, 0xc0, 0x72, 0x00, 0x01, 0x00, 0x01, 0x00, 0x00, 0x00,
    0x23, 0x00, 0x04, 0x36, 0x95, 0x5e, 0x45
};
~~~

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
htibosch: > If @htibosch and @gkwicker are okay with this, I have no further objection :)

I am OK with the change. I do not have an overview of all available releases of `stdint.h`, but my feeling is that `intptr_t` is always part of it.
`uintptr_t` would be **better** though: pointers are also kind of unsigned, with `NULL` at the bottom.

( apart from this PR: there is more work to do on pointers: in every packet buffer space (`pucEthernetBuffer`), there is a hidden pointer, pointing to the owning `NetworkBufferDescriptor_t`.
This would also be a good place to adopt `uintptr_t`, in stead of just casting to `uint32_t` )

htibosch:
Just to be sure, I ran the same test on a big-endian platform. The results were exactly the same:
~~~c
LITTLE endian STM32F4
Swap = 0 CRC = 224C 442A 224C 442A
Swap = 1 CRC = 224C 224C 224C 224C

BIG endian Atmel UC3A
Swap = 0 CRC = 224C 442A 224C 442A
Swap = 1 CRC = 224C 224C 224C 224C
~~~

andyjpb:
I have resubmitted this as https://github.com/FreeRTOS/FreeRTOS/pull/286.

htibosch:
> 
> 
> I have resubmitted this as #286.

Thanks a lot for the new PR Andy. You can now close this PR #122. Thanks, Hein

andyjpb:
Hi,

> I have resubmitted this as #286.
>
> Thanks a lot for the new PR Andy. You can now close this PR 
> #122. Thanks, Hein

Done!



Best wishes,
@ndy

-- 
andyjpb@ashurst.eu.org
http://www.ashurst.eu.org/
0x7EBA75FF

PR #123, opened 2020-07-04
Try harder to avoid DHCP race condition and potential live-lock

andyjpb:
There is a race condition in the stack between the IPTask and the Ethernet Driver Task.

Description
-----------
I'm not sure that this is the best way to "fix" the race. Technically the race still exists and if the IPTask doesn't get scheduled for
more than 2,500ms then the condition will reappear.

Perhaps it's better to rewrite vDHCPProcess()'s eLeasedAddress state to check whether the DHCP lease really is up for renewal?

Anyway, this is a patch for discussion and makes the issue go away for me for now.

Test Steps
-----------
This is a race condition so is inherently difficult to reproduce.
I discovered it because I had some FreeRTOS_debug_printfs()'s scattered around and was running packet captures on my development machine.

Sometimes when requesting a DHCP lease the stack will get stuck in a loop of requesting acknowledging and re-requesting. If this happens you will see a lot of network traffic from the target and repeated Network Up notifications.

### DIFF
--- a/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_DHCP.c
+++ b/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_DHCP.c
@@ -58,7 +58,7 @@
 
 /* Timer parameters */
 #ifndef dhcpINITIAL_DHCP_TX_PERIOD
-	#define dhcpINITIAL_TIMER_PERIOD			( pdMS_TO_TICKS( 250U ) )
+	#define dhcpINITIAL_TIMER_PERIOD			( pdMS_TO_TICKS( 2500U ) )
 	#define dhcpINITIAL_DHCP_TX_PERIOD			( pdMS_TO_TICKS( 5000U ) )
 #endif
 


htibosch:
@andyjpb, I am truly sorry to see that this PR was "forgotten". It is good that you sent me a ping to remind me.
I was able to replicate the problem and I found a solution for it.  I will come back to it.
Thanks for reporting, Hein

AniruddhaKanhere:
Hello @andyjpb, thank you for being a contributor to the FreeRTOS community. 
Can you please recreate this PR in the new FreeRTOS-Plus-TCP repository [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP).
The FreeRTOS+TCP source code does not reside in FreeRTOS/FreeRTOS (this) repository anymore.

Apologies for the inconvenience.

AniruddhaKanhere:
This issue was fixed with PR https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/62.
Thanks @andyjpb for pointing out this issue and taking the time to provide a fix. I am closing this PR. Please feel free to re-open this PR or a new PR/issue in the [FreeRTOS+TCP](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP) repository if the problem still persists.

PR #119, opened 2020-07-02
MQTT Lightweight demo

leegeth:
MQTT Lightweight demo

Description
-----------
Demo for showing use of the MQTT light weight serializer API.
The Light weight serializer API lets user to serialize and
deserialize MQTT messages into a user provided buffer.
This API allows use of a statically allocated buffer.

The Example shown below uses this API to create MQTT messages and
send them over the connection established using FreeRTOS sockets.
The example is single threaded and uses statically allocated memory;
it uses QOS0 and therefore does not implement any retransmission
mechanism for Publish messages.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #132, opened 2020-07-07
Add demo completion string for CI checks to pass

aggarw13:
Add success string in stubbed Lightweight MQTT demo to support CI nightly checks 

**Note**: The complete implementation for the Beta2 light-weight MQTT demo is being added in #119 


muneebahmed10:
Looks like this PR is no longer necessary since the demo has been implemented

PR #142, opened 2020-07-14
Update plain text demo to use local broker

leegeth:
Update plain text demo to use local broker

Description
-----------
The change includes
1. Update plain text demo to use local broker.
2. Update the transport implementation to set the socket options for receive timeout after a connect is successful. This is needed as the local connection is failing if reversed the order. Will be following up with TCP team for this issue separately.

Test Steps
-----------
Ran the light weight demo and plain text demo successfully.

### DIFF
--- a/FreeRTOS-Plus/Source/FreeRTOS-IoT-Libraries-LTS-Beta2/abstractions/platform/freertos/transport_interface_freertos.c
+++ b/FreeRTOS-Plus/Source/FreeRTOS-IoT-Libraries-LTS-Beta2/abstractions/platform/freertos/transport_interface_freertos.c
@@ -57,20 +57,6 @@ BaseType_t Transport_FreeRTOS_Connect( NetworkContext_t * pNetworkContext,
         LogError( ( "Failed to create new socket." ) );
         socketStatus = -1;
     }
-    else
-    {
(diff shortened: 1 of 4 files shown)


FreeRTOS-Plus/Source/FreeRTOS-IoT-Libraries-LTS-Beta2/abstractions/platform/freertos/transport_interface_freertos.c
```diff
@@ -91,6 +77,27 @@ BaseType_t Transport_FreeRTOS_Connect( NetworkContext_t * pNetworkContext,
...
+        {
+            LogError( ( "Failed to connect to %s. FreeRTOS_connect returned with status=%d.",
+                        pHostName,
+                        socketStatus ) );
+        }
+    }
+
+    if( socketStatus == 0 )
+    {
+        socketStatus = FreeRTOS_setsockopt( tcpSocket,
```
abhidixi11, line 91: I think we should  add send timeout as well.

leegeth: Yea agree. I think that will be a good change to be in a separate PR.

PR #146, opened 2020-07-14
Update main.c logs to use logging stack

leegeth:
Description
-----------
Update main.c logs to use logging stack.

### DIFF
--- a/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/common/main.c
+++ b/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/common/main.c
@@ -45,8 +45,12 @@
 #include "FreeRTOS_IP.h"
 #include "FreeRTOS_Sockets.h"
 
-/* Demo app includes. */
+/* Demo logging includes. */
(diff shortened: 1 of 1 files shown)


FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/common/main.c
```diff
@@ -45,9 +45,13 @@
 #include "FreeRTOS_IP.h"
 #include "FreeRTOS_Sockets.h"
 
-/* Demo app includes. */
+/* Demo logging includes. */
 #include "demo_logging.h"
```
aggarw13, line 49: Is this include required or should the `demo_config.h` include this header file?

leegeth: This include is required to call the logging init, otherwise a compile time warning is observed.

leegeth: logging init is called only from main.c, I think that way it is better to be only included in here.

PR #148, opened 2020-07-15
Network interface zynq frametype

htibosch:
Description
-----------
The EMAC of the Zynq 7000 can only calculate the checksum for TCP and UDP, and not for ICMP, IGMP.
So also when `ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM` is defined, the function `xNetworkInterfaceOutput()` will calculate the checksum for ICMP packets.
By mistake, `usGenerateProtocolChecksum()` would also be called for frame-types different from IPv4: ARP.
This PR will correct that behaviour.

Without this patch, the ARP will also work properly, but ARP packets should not be passed to `usGenerateProtocolChecksum()` of course.

The proposed change:
~~~c
-    if( ( pxPacket->xICMPPacket.xIPHeader.ucProtocol != ipPROTOCOL_UDP ) &&
+    if( ( pxPacket->xICMPPacket.xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
+        ( pxPacket->xICMPPacket.xIPHeader.ucProtocol != ipPROTOCOL_UDP ) &&
         ( pxPacket->xICMPPacket.xIPHeader.ucProtocol != ipPROTOCOL_TCP ) )
     {
         /* The EMAC will calculate the checksum of the IP-header.
         It can only calculate protocol checksums of UDP and TCP,
         so for ICMP and other protocols it must be done manually. */
         usGenerateProtocolChecksum( (uint8_t*)&( pxPacket->xUDPPacket ), pxBuffer->xDataLength, pdTRUE );
     }
~~~

Test Steps
-----------
Enabled the logging macro `FreeRTOS_printf()`.
Connect or ping from the Zynq to an external device.
See if the following length check error still appears in the logging:

    CRC error: 1234 location 3

I tested all 3 protocols: ICMP, UDP and TCP.


htibosch:
I have submitted a new fresh PR #177 that will replace this one.
The reason is that I want to recreate my personal fork of FreeRTOS/FreeRTOS.
Hein

PR #152, opened 2020-07-17
Update URL for new MQTT demo

muneebahmed10:
Description
-----------
Updates the demo URL for the new MQTT library

### DIFF
--- a/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta1/readme.txt
+++ b/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta1/readme.txt
@@ -30,6 +30,7 @@ libraries:
   + https://www.FreeRTOS.org/jobs/
   + https://www.FreeRTOS.org/ota/
   + https://www.FreeRTOS.org/pkcs11/
+  + https://www.FreeRTOS.org/mqtt_lts/
 
 
 *** LOCATING THE EXAMPLE PROJECTS ***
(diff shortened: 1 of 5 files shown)


FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/common/main.c
```diff
@@ -26,7 +26,7 @@
  */
 
 /***
- * See https://www.FreeRTOS.org/mqtt/index.html for configuration and usage instructions.
+ * See https://www.FreeRTOS.org/mqtt_lts/index.html for configuration and usage instructions.
```
leegeth, line 29: There are a few places  we would need to update this .url file, coupl[e of readmes I think. Can you please update those as well?

muneebahmed10: Updated

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta1/readme.txt
```diff
@@ -30,6 +30,7 @@ libraries:
   + https://www.FreeRTOS.org/jobs/
   + https://www.FreeRTOS.org/ota/
   + https://www.FreeRTOS.org/pkcs11/
+  + https://www.FreeRTOS.org/mqtt_lts/
```
leegeth, line 33: Did we need it here? This seems to contain the demo for Beta1 libraries only

muneebahmed10: Good catch, I guess this what happens when the directories differ by 1 character

PR #153, opened 2020-07-17
Add mqtt lts URL to all READMEs

muneebahmed10:
Description
-----------
Update the URL again

### DIFF
--- a/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta1/readme.txt
+++ b/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta1/readme.txt
@@ -30,6 +30,7 @@ libraries:
   + https://www.FreeRTOS.org/jobs/
   + https://www.FreeRTOS.org/ota/
   + https://www.FreeRTOS.org/pkcs11/
+  + https://www.FreeRTOS.org/mqtt_lts/
 
 
 *** LOCATING THE EXAMPLE PROJECTS ***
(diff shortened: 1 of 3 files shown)


FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta1/readme.txt
```diff
@@ -30,7 +30,6 @@ libraries:
   + https://www.FreeRTOS.org/jobs/
   + https://www.FreeRTOS.org/ota/
   + https://www.FreeRTOS.org/pkcs11/
-  + https://www.FreeRTOS.org/mqtt_lts/
```
aggarw13, line 33: Reason for removing from Beta1 readme.txt?

The same file also mentions about the Beta2 demos

abhidixi11: It was not there before as far as I know. I got added in the earlier PR.

muneebahmed10: I did not mean to add it to this file in #152. However, if it's meant to be here too, then I can push an update to keep it

muneebahmed10: Added it back. This way the READMEs will be identical, as they were before #152

PR #156, opened 2020-07-21
Add Atmel Studio projects for ATMega4809 and AVR128DA48

m17336:
Added Atmel Studio projects for ATMega4809 and AVR128DA48

Description
-----------
Beside the projects there are a couple of changes in Demo/Common files. These changes are:
o	(1) In semtest.h I updated the return type of xAreSemaphoreTasksStillRunning function from BaseType_t to portBASE_TYPE because this is the return type it has where it is implemented (semtest.c).
o	(2) In recmutex.c I added an explicit cast to UBaseType_t On line 260. I did this because the right-hand expression in the “==” check was, at some point, expected to roll over and therefore be equal to the left-hand side (that was, at that time, 0). Instead, what I think was happening was that the compiler performed integer promotion and turned the 8 bit variable into a 32 bit one, this way roll over did not happened and it was actually comparing 0 to 256 instead of 0 to 255 + 1 (=0 because of roll over).

Test Steps
-----------
From what I can tell the above changes should be backwards compatible because:  in change (1) the two types actually resolve to the same primitive type and in change (2) the cast actually serves as an additional clarification to the "compiler" about what the code is expected to do.
But you might have a bigger picture than me, and have some more tests to perform in order to confirm the changes are ok, so this is way I described them in detail.

### DIFF
--- a/FreeRTOS/Test/VeriFast/Makefile
+++ b/FreeRTOS/Test/VeriFast/Makefile
@@ -18,35 +18,35 @@ all: queue
 
 .PHONY: queue
 queue:
-	@$(VERIFAST) $(VERIFAST_ARGS)                         queue/create.c                    | $(call check_coverage,269)
-	@$(VERIFAST) $(VERIFAST_ARGS)                         queue/prvCopyDataFromQueue.c      | $(call check_coverage,253)
-	@$(VERIFAST) $(VERIFAST_ARGS) -disable_overflow_check queue/prvCopyDataToQueue.c        | $(call check_coverage,280)
-	@$(VERIFAST) $(VERIFAST_ARGS)                         queue/prvIsQueueEmpty.c           | $(call check_coverage,234)
-	@$(VERIFAST) $(VERIFAST_ARGS)                         queue/prvIsQueueFull.c            | $(call check_coverage,234)
-	@$(VERIFAST) $(VERIFAST_ARGS)                         queue/prvLockQueue.c              | $(call check_coverage,235)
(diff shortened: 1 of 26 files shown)


FreeRTOS/Demo/Common/include/semtest.h
```diff
@@ -29,7 +29,7 @@
 #define SEMAPHORE_TEST_H
 
 void vStartSemaphoreTasks( UBaseType_t uxPriority );
-BaseType_t xAreSemaphoreTasksStillRunning( void );
```
yuhui-zheng, line 32: Thank you for fixing this inconsistency. 

I may want to update your PR to use ```BaseType_t``` in both ```semtest.{c,h}```. Since according to https://www.freertos.org/upgrading-to-FreeRTOS-V8.html, ```portBASE_TYPE``` is the old name, and ```BaseType_t``` is the new name (since FreeRTOS V8.x.x ).

FreeRTOS/Demo/Common/Minimal/recmutex.c
```diff
@@ -257,7 +257,7 @@ static void prvRecursiveMutexBlockingTask( void *pvParameters )
 		}
 
 		/* The controlling and blocking tasks should be in lock step. */
-		if( uxControllingCycles != ( uxBlockingCycles + 1 ) )
```
yuhui-zheng, line 260: Acknowledged. Thanks.

yuhui-zheng:
Oops... By rebasing, force pushing source branch I accidentally closed the PR and now it won't let me reopen... will figure this out.

Issue #136, opened 2020-07-10
[Inquiry] A missing error assignment in QPeek.c at the Minimal folder


kaizsv:
Hello, I found that an error assignment might be missed in QPeek.c at the Minimal test folder.

https://github.com/FreeRTOS/FreeRTOS/blob/d7667a0034841f2968f9f9f805030cc608bfbce1/FreeRTOS/Demo/Common/Minimal/QPeek.c#L367

lundinc2:
@kaizsv,

Thanks for bringing this to our attention, I will investigate this.

lundinc2:
I have now merged the above PR, thank you again for this report.

PR #166, opened 2020-07-24
Add missing error state assignment

lundinc2:
Add missing error state assignment

Description
-----------
QPeek was not setting the error state in one the if statements

Test Steps
-----------

Related Issue
-----------
https://github.com/FreeRTOS/FreeRTOS/issues/136


lundinc2:
This merely a proposed fix, need to do further testing.

PR #168, opened 2020-07-26
Use new QEMU test project to improve stream/message buffer tests

RichardBarry:
Description
-----------
Improve the robustness of the stream and message buffer tests - required various minor updates to StreamBufferDemo.c and MessageBufferDemo.c to account for variations in the execution pattern related to how far through a tick period tasks started executing and the priority of other test tasks.

Also check in the FreeRTOS/Demo/CORTEX_LM3S6965_QEMU project used to first investigate then update the StreamBufferDemo.c and Message BufferDemo.c files.  The QEMU project is useful for recording execution traces, using thread aware FreeRTOS plug-ins, and testing different GCC versions.

Test Steps
-----------
Run the CORTEX_LM3S6965_QEMU project as will be documented on the FreeRTOS.org site.  Leave the project running and observe the QEMU output.  Note the demo will eventually fail due to too many tick interrupts occurring between the the task that tests the stream buffer trigger level unblocking and subsequently executing - that failure does not occur on real hardware.

Related Issue
-----------
None.


yuhui-zheng:
We'll likely need to add some user oriented instructions. Meanwhile, leaving a note for none Windows users.

These minor changes may be necessary -- 
- ```chmod +x StartQEMU.bat```
- Watch out for the debug folder name. In my setup StartQEMU.bat has ```qemu-system-arm -machine lm3s6965evb -s -S -kernel ./Debug/RTOSDemo.elf```
- Correspondingly ```RTOSDemo Debug.launch``` has ```<stringAttribute key="org.eclipse.cdt.launch.PROGRAM_NAME" value="Debug/RTOSDemo.elf"/>```

A clean Eclipse installation may need to setup path for QEMU and ARM toolchain as well. 

I'm hitting "error in STRM" after a while, though per previous communication this is likely in QEMU port as we don't see it in hardware. We'll need to look into QEMU as a separate effort to this PR.

PR #176, opened 2020-07-29
MQTT mutual auth demo.

leegeth:
MQTT mutual auth demo.

Description
-----------
MQTT mutual auth demo.

Demo for showing use of the managed MQTT API over a mutually authenticated connection.

The Example shown below uses this API to create MQTT messages and
send them over the connection established using FreeRTOS sockets.
The example is single threaded and uses statically allocated memory;
it uses QoS1.

A mutually authenticated TLS connection is used to connect to the AWS IoT
MQTT message broker in this example. Define democonfigAWS_ROOT_CA_PEM,
democonfigCLIENT_CERTIFICATE_PEM, and democonfigCLIENT_PRIVATE_KEY_PEM in
demo_config.h to achieve mutual authentication.

Test Steps
-----------
1. Ran the demo.

### DIFF
--- a/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
+++ b/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
@@ -140,8 +140,8 @@
 #define mqttexampleKEEP_ALIVE_TIMEOUT_SECONDS             ( 60U )
 
 /**
- * @brief Delay between consecutive cycles of MQTT publish operations in a
- * demo iteration in ticks.
+ * @brief Delay (in ticks) between consecutive cycles of MQTT publish operations in a
+ * demo iteration.
  *
  * Note that the process loop also has a timeout, so the total time between
  * publishes is the sum of the two delays.
@@ -340,9 +340,10 @@ static void prvMQTTDemoTask( void * pvParameters )
     /* Remove compiler warnings about unused parameters. */
     ( void ) pvParameters;
 
-    /* Store the entry time of the demo application. This entry time is used for
-     * calculating the elapsed time in application, by the time utility function
-     * used by MQTT library. */
+    /* Set the entry time of the demo application. This entry time will be used
+     * to calculate relative time elapsed in the execution of the demo application,
+     * by the timer utility function that is provided to the MQTT library.
+     */
     ulGlobalEntryTimeMs = prvGetTimeMs();
 
     for( ; ; )
@@ -370,7 +371,7 @@ static void prvMQTTDemoTask( void * pvParameters )
          * The function #prvMQTTSubscribeToTopic will not wait to receive a SUBACK,
          * but the function #MQTT_ProcessLoop will attempt to receive the SUBACK
          * from network and if a SUBACK is received, application will be notified
-         * through the callback registered(#prvEventCallback for this application).
+         * through the callback registered (#prvEventCallback for this application).
          * This demo uses QoS1 in Subscribe, therefore, the Publish messages
          * received from the broker will have QoS1. */
         LogInfo( ( "Attempt to subscribe to the MQTT topic %s.\r\n", mqttexampleTOPIC ) );
--- a/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
+++ b/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
@@ -213,9 +213,9 @@ static void prvMQTTUnsubscribeFromTopic( MQTTContext_t * pxMQTTContext );
 static uint32_t prvGetTimeMs( void );
 
 /**
- * @brief Process a response or ack to an MQTT request (PING, SUBSCRIBE
- * or UNSUBSCRIBE). This function processes PINGRESP, PUBACK, SUBACK, and
- * UNSUBACK.
+ * @brief Process a response or ack to an MQTT request (PING, PUBLISH,
+ * SUBSCRIBE or UNSUBSCRIBE). This function processes PINGRESP, PUBACK,
+ * SUBACK, and UNSUBACK.
  *
  * @param[in] pxIncomingPacket is a pointer to structure containing deserialized
  * MQTT response.
@@ -467,8 +467,6 @@ static void prvCreateMQTTConnectionWithBroker( MQTTContext_t * pxMQTTContext,
     MQTTConnectInfo_t xConnectInfo;
     bool xSessionPresent;
     TransportInterface_t xTransport;
-    MQTTGetCurrentTimeFunc_t xGetTimeFunction;
-    MQTTEventCallback_t xUserCallback;
 
     /***
      * For readability, error handling in this function is restricted to the use of
@@ -480,15 +478,8 @@ static void prvCreateMQTTConnectionWithBroker( MQTTContext_t * pxMQTTContext,
     xTransport.send = TLS_FreeRTOS_send;
     xTransport.recv = TLS_FreeRTOS_recv;
 
-    /* Application callback for receiving incoming published and incoming acks
-     * from MQTT library. */
-    xUserCallback = prvEventCallback;
-
-    /* Time utility function to be used by MQTT library. */
-    xGetTimeFunction = prvGetTimeMs;
-
     /* Initialize MQTT library. */
-    xResult = MQTT_Init( pxMQTTContext, &xTransport, xGetTimeFunction, xUserCallback, &xBuffer );
+    xResult = MQTT_Init( pxMQTTContext, &xTransport, prvGetTimeMs, prvEventCallback, &xBuffer );
     configASSERT( xResult == MQTTSuccess );
 
     /* Many fields not used in this demo so start with everything at 0. */
--- a/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
+++ b/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
@@ -58,7 +58,7 @@
 #include "mqtt.h"
 
 /* Transport interface include. */
-#include "mbedtls_freertos.h"
+#include "tls_freertos.h"
 
 /*-----------------------------------------------------------*/
 
@@ -400,7 +400,7 @@ static void prvMQTTDemoTask( void * pvParameters )
         MQTT_Disconnect( &xMQTTContext );
 
         /* Close the network connection.  */
-        Mbedtls_FreeRTOS_Disconnect( &xNetworkContext );
+        TLS_FreeRTOS_Disconnect( &xNetworkContext );
 
         /* Wait for some time between two iterations to ensure that we do not
          * bombard the public test mosquitto broker. */
@@ -418,21 +418,21 @@ static void prvTLSConnect( NetworkCredentials_t * pxNetworkCredentials,
     BaseType_t xNetworkStatus;
 
     /* Set the credentials for establishing a TLS connection. */
-    pxNetworkCredentials->pRootCa = democonfigAWS_ROOT_CA_PEM;
+    pxNetworkCredentials->pRootCa = ( const unsigned char * ) democonfigAWS_ROOT_CA_PEM;
     pxNetworkCredentials->rootCaSize = sizeof( democonfigAWS_ROOT_CA_PEM );
-    pxNetworkCredentials->pClientCert = democonfigCLIENT_CERTIFICATE_PEM;
+    pxNetworkCredentials->pClientCert = ( const unsigned char * ) democonfigCLIENT_CERTIFICATE_PEM;
     pxNetworkCredentials->clientCertSize = sizeof( democonfigCLIENT_CERTIFICATE_PEM );
-    pxNetworkCredentials->pPrivateKey = democonfigCLIENT_PRIVATE_KEY_PEM;
+    pxNetworkCredentials->pPrivateKey = ( const unsigned char * ) democonfigCLIENT_PRIVATE_KEY_PEM;
     pxNetworkCredentials->privateKeySize = sizeof( democonfigCLIENT_PRIVATE_KEY_PEM );
 
     /* Attempt to create a mutually authenticated TLS connection. */
-    xNetworkStatus = Mbedtls_FreeRTOS_Connect( pxNetworkContext,
-                                               democonfigAWS_IOT_ENDPOINT,
-                                               democonfigMQTT_BROKER_PORT,
-                                               pxNetworkCredentials,
-                                               TRANSPORT_SEND_RECV_TIMEOUT_MS,
-                                               TRANSPORT_SEND_RECV_TIMEOUT_MS );
-    configASSERT( xNetworkStatus == MBEDTLS_SUCCESS );
+    xNetworkStatus = TLS_FreeRTOS_Connect( pxNetworkContext,
+                                           democonfigAWS_IOT_ENDPOINT,
+                                           democonfigMQTT_BROKER_PORT,
+                                           pxNetworkCredentials,
+                                           TRANSPORT_SEND_RECV_TIMEOUT_MS,
+                                           TRANSPORT_SEND_RECV_TIMEOUT_MS );
+    configASSERT( xNetworkStatus == TLS_TRANSPORT_SUCCESS );
     LogInfo( ( "A mutually authenticated TLS connection established with %s.\r\n",
                democonfigAWS_IOT_ENDPOINT ) );
 }
@@ -453,8 +453,8 @@ static void prvCreateMQTTConnectionWithBroker( MQTTContext_t * pxMQTTContext,
 
     /* Fill in Transport Interface send and receive function pointers. */
     xTransport.pNetworkContext = pxNetworkContext;
-    xTransport.send = Mbedtls_FreeRTOS_send;
-    xTransport.recv = Mbedtls_FreeRTOS_recv;
+    xTransport.send = TLS_FreeRTOS_send;
+    xTransport.recv = TLS_FreeRTOS_recv;
 
     /* Application callbacks for receiving incoming published and incoming acks
      * from MQTT library. */
--- a/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
+++ b/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
@@ -332,7 +332,7 @@ static void prvMQTTDemoTask( void * pvParameters )
     {
         /****************************** Connect. ******************************/
 
-        /* Establish a TCP connection with the MQTT broker. This example connects to
+        /* Establish a TLS connection with the MQTT broker. This example connects to
          * the MQTT broker as specified in democonfigAWS_IOT_ENDPOINT and
          * democonfigMQTT_BROKER_PORT at the top of this file. */
         LogInfo( ( "Creating a TLS connection to %s.\r\n", democonfigAWS_IOT_ENDPOINT ) );
--- a/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
+++ b/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
@@ -1,0 +1,697 @@
+/*
+ * FreeRTOS Kernel V10.3.0
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * http://www.FreeRTOS.org
+ * http://aws.amazon.com/freertos
+ *
+ * 1 tab == 4 spaces!
+ */
+
+/*
+ * Demo for showing use of the managed MQTT API.
+ *
+ * The Example shown below uses this API to create MQTT messages and
+ * send them over the connection established using FreeRTOS sockets.
+ * The example is single threaded and uses statically allocated memory;
+ * it uses QoS1.
+ *
+ * !!! NOTE !!!
+ * This MQTT demo does not authenticate the server nor the client.
+ * Hence, this demo should not be used as production ready code.
+ */
+
+/* Standard includes. */
+#include <string.h>
+#include <stdio.h>
+
+/* Kernel includes. */
+#include "FreeRTOS.h"
+#include "task.h"
+
+/* FreeRTOS+TCP includes. */
+#include "FreeRTOS_IP.h"
+#include "FreeRTOS_Sockets.h"
+
+/* Demo Specific configs. */
+#include "demo_config.h"
+
+/* MQTT library includes. */
+#include "mqtt.h"
+
+/* Transport interface include. */
+#include "mbedtls_freertos.h"
+
+/*-----------------------------------------------------------*/
+
+/* Compile time error for undefined configs. */
+#ifndef democonfigAWS_IOT_ENDPOINT
+    #error "Define the config democonfigAWS_IOT_ENDPOINT by following the instructions in file demo_config.h."
+#endif
+#ifndef democonfigAWS_ROOT_CA_PEM
+    #error "Please define Root CA certificate of the MQTT broker(democonfigAWS_ROOT_CA_PEM) in demo_config.h."
+#endif
+#ifndef democonfigCLIENT_CERTIFICATE_PEM
+    #error "Please define client certificate(democonfigCLIENT_CERTIFICATE_PEM) in demo_config.h."
+#endif
+#ifndef democonfigCLIENT_PRIVATE_KEY_PEM
+    #error "Please define client private key(democonfigCLIENT_PRIVATE_KEY_PEM) in demo_config.h."
+#endif
+
+/*-----------------------------------------------------------*/
+
+/* Default values for configs. */
+#ifndef democonfigCLIENT_IDENTIFIER
+
+/**
+ * @brief The MQTT client identifier used in this example.  Each client identifier
+ * must be unique so edit as required to ensure no two clients connecting to the
+ * same broker use the same client identifier.
+ *
+ * @note Appending __TIME__ to the client id string will reduce the possibility of a
+ * client id collision in the broker. Note that the appended time is the compilation
+ * time. This client id can cause collision, if more than one instance of the same
+ * binary is used at the same time to connect to the broker.
+ */
+    #define democonfigCLIENT_IDENTIFIER    "testClient"__TIME__
+#endif
+
+#ifndef democonfigMQTT_BROKER_PORT
+
+/**
+ * @brief The port to use for the demo.
+ */
+    #define democonfigMQTT_BROKER_PORT    ( 8883 )
+#endif
+
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Timeout for receiving CONNACK packet in milliseconds.
+ */
+#define mqttexampleCONNACK_RECV_TIMEOUT_MS          ( 1000U )
+
+/**
+ * @brief The topic to subscribe and publish to in the example.
+ *
+ * The topic name starts with the client identifier to ensure that each demo
+ * interacts with a unique topic name.
+ */
+#define mqttexampleTOPIC                            democonfigCLIENT_IDENTIFIER "/example/topic"
+
+/**
+ * @brief The MQTT message published in this example.
+ */
+#define mqttexampleMESSAGE                          "Hello World!"
+
+/**
+ * @brief Dimensions a file scope buffer currently used to send and receive MQTT data
+ * from a socket.
+ */
+#define mqttexampleSHARED_BUFFER_SIZE               ( 500U )
+
+/**
+ * @brief Time to wait between each cycle of the demo implemented by prvMQTTDemoTask().
+ */
+#define mqttexampleDELAY_BETWEEN_DEMO_ITERATIONS    ( pdMS_TO_TICKS( 5000U ) )
+
+/**
+ * @brief Timeout for MQTT_ProcessLoop in milliseconds.
+ */
+#define mqttexamplePROCESS_LOOP_TIMEOUT_MS          ( 500U )
+
+/**
+ * @brief Keep alive time reported to the broker while establishing an MQTT connection.
+ *
+ * It is the responsibility of the Client to ensure that the interval between
+ * Control Packets being sent does not exceed the this Keep Alive value. In the
+ * absence of sending any other Control Packets, the Client MUST send a
+ * PINGREQ Packet.
+ */
+#define mqttexampleKEEP_ALIVE_TIMEOUT_SECONDS       ( 60U )
+
+/**
+ * @brief Delay between MQTT publishes. Note that the process loop also has a
+ * timeout, so the total time between publishes is the sum of the two delays.
+ */
+#define mqttexampleDELAY_BETWEEN_PUBLISHES          ( pdMS_TO_TICKS( 500U ) )
+
+/**
+ * @brief Transport timeout in milliseconds for transport send and receive.
+ */
+#define TRANSPORT_SEND_RECV_TIMEOUT_MS              ( 200U )
+
+#define _MILLISECONDS_PER_SECOND                    ( 1000U )                                         /**< @brief Milliseconds per second. */
+#define _MILLISECONDS_PER_TICK                      ( _MILLISECONDS_PER_SECOND / configTICK_RATE_HZ ) /**< Milliseconds per FreeRTOS tick. */
+
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief The task used to demonstrate the MQTT API.
+ *
+ * @param[in] pvParameters Parameters as passed at the time of task creation. Not
+ * used in this example.
+ */
+static void prvMQTTDemoTask( void * pvParameters );
+
+/**
+ * @brief Sends an MQTT Connect packet over the already connected TCP socket.
+ *
+ * @param pxMQTTContext MQTT context pointer.
+ * @param xNetworkContext network context.
+ *
+ */
+static void prvCreateMQTTConnectionWithBroker( MQTTContext_t * pxMQTTContext,
+                                               NetworkContext_t * pxNetworkContext );
+
+/**
+ * @brief Subscribes to the topic as specified in mqttexampleTOPIC at the top of
+ * this file.
+ *
+ * @param pxMQTTContext MQTT context pointer.
+ */
+static void prvMQTTSubscribeToTopic( MQTTContext_t * pxMQTTContext );
+
+/**
+ * @brief  Publishes a message mqttexampleMESSAGE on mqttexampleTOPIC topic.
+ *
+ * @param pxMQTTContext MQTT context pointer.
+ */
+static void prvMQTTPublishToTopic( MQTTContext_t * pxMQTTContext );
+
+/**
+ * @brief Unsubscribes from the previously subscribed topic as specified
+ * in mqttexampleTOPIC.
+ *
+ * @param pxMQTTContext MQTT context pointer.
+ */
+static void prvMQTTUnsubscribeFromTopic( MQTTContext_t * pxMQTTContext );
+
+/**
+ * @brief The timer query function provided to the MQTT context.
+ *
+ * @return Time in milliseconds.
+ */
+static uint32_t prvGetTimeMs( void );
+
+/**
+ * @brief Process a response or ack to an MQTT request (PING, SUBSCRIBE
+ * or UNSUBSCRIBE). This function processes PINGRESP, SUBACK, UNSUBACK
+ *
+ * @param pxIncomingPacket is a pointer to structure containing deserialized
+ * MQTT response.
+ * @param usPacketId is the packet identifier from the ack received.
+ */
+static void prvMQTTProcessResponse( MQTTPacketInfo_t * pxIncomingPacket,
+                                    uint16_t usPacketId );
+
+/**
+ * @brief Process incoming Publish message.
+ *
+ * @param pxPublishInfo is a pointer to structure containing deserialized
+ * Publish message.
+ */
+static void prvMQTTProcessIncomingPublish( MQTTPublishInfo_t * pxPublishInfo );
+
+/**
+ * @brief The application callback function for getting the incoming publish
+ * and incoming acks reported from the MQTT library.
+ *
+ * @param pxMQTTContext MQTT context pointer.
+ * @param pxPacketInfo Packet Info pointer for the incoming packet.
+ * @param usPacketIdentifier Packet identifier of the incoming packet.
+ * @param pxPublishInfo Deserialized publish info pointer for the incoming
+ * packet.
+ */
+static void prvEventCallback( MQTTContext_t * pxMQTTContext,
+                              MQTTPacketInfo_t * pxPacketInfo,
+                              uint16_t usPacketIdentifier,
+                              MQTTPublishInfo_t * pxPublishInfo );
+
+/**
+ * @brief TLS connect to endpoint democonfigAWS_IOT_ENDPOINT.
+ *
+ * @param pxNetworkCredentials Network credentials to establish a TLS connection
+ * with democonfigAWS_IOT_ENDPOINT.
+ * @param pxNetworkCredentials Network context.
+ */
+static void prvTLSConnect( NetworkCredentials_t * pxNetworkCredentials,
+                           NetworkContext_t * pxNetworkContext );
+
+/*-----------------------------------------------------------*/
+
+/* @brief Static buffer used to hold MQTT messages being sent and received. */
+static uint8_t ucSharedBuffer[ mqttexampleSHARED_BUFFER_SIZE ];
+
+/**
+ * @brief Global entry time into the application to use as a reference timestamp
+ * in the #prvGetTimeMs function. #prvGetTimeMs will always return the difference
+ * between the current time and the global entry time. This will reduce the chances
+ * of overflow for the 32 bit unsigned integer used for holding the timestamp.
+ */
+static uint32_t ulGlobalEntryTimeMs;
+
+/**
+ * @brief Packet Identifier generated when Publish request was sent to the broker;
+ * it is used to match received Publish ACK to the transmitted ACK.
+ */
+static uint16_t usPublishPacketIdentifier;
+
+/**
+ * @brief Packet Identifier generated when Subscribe request was sent to the broker;
+ * it is used to match received Subscribe ACK to the transmitted ACK.
+ */
+static uint16_t usSubscribePacketIdentifier;
+
+/**
+ * @brief Packet Identifier generated when Unsubscribe request was sent to the broker;
+ * it is used to match received Unsubscribe response to the transmitted unsubscribe
+ * request.
+ */
+static uint16_t usUnsubscribePacketIdentifier;
+
+
+/** @brief Static buffer used to hold MQTT messages being sent and received. */
+static MQTTFixedBuffer_t xBuffer =
+{
+    .pBuffer = ucSharedBuffer,
+    .size    = mqttexampleSHARED_BUFFER_SIZE
+};
+
+/*-----------------------------------------------------------*/
+
+/*
+ * @brief Create the task that demonstrates the Plain text MQTT API Demo.
+ */
+void vStartSimpleMQTTDemo( void )
+{
+    /* This example uses a single application task, which in turn is used to
+     * connect, subscribe, publish, unsubscribe and disconnect from the MQTT
+     * broker. */
+    xTaskCreate( prvMQTTDemoTask,          /* Function that implements the task. */
+                 "MQTTDemo",               /* Text name for the task - only used for debugging. */
+                 democonfigDEMO_STACKSIZE, /* Size of stack (in words, not bytes) to allocate for the task. */
+                 NULL,                     /* Task parameter - not used in this case. */
+                 tskIDLE_PRIORITY,         /* Task priority, must be between 0 and configMAX_PRIORITIES - 1. */
+                 NULL );                   /* Used to pass out a handle to the created task - not used in this case. */
+}
+/*-----------------------------------------------------------*/
+
+static void prvMQTTDemoTask( void * pvParameters )
+{
+    uint32_t ulPublishCount = 0U;
+    const uint32_t ulMaxPublishCount = 5UL;
+    NetworkContext_t xNetworkContext = { 0 };
+    NetworkCredentials_t xNetworkCredentials = { 0 };
+    MQTTContext_t xMQTTContext;
+    MQTTStatus_t xMQTTStatus;
+
+    /* Remove compiler warnings about unused parameters. */
+    ( void ) pvParameters;
+
+    ulGlobalEntryTimeMs = prvGetTimeMs();
+
+    for( ; ; )
+    {
+        /****************************** Connect. ******************************/
+
+        /* Establish a TCP connection with the MQTT broker. This example connects to
+         * the MQTT broker as specified in democonfigAWS_IOT_ENDPOINT and
+         * democonfigMQTT_BROKER_PORT at the top of this file. */
+        LogInfo( ( "Creating a TLS connection to %s.\r\n", democonfigAWS_IOT_ENDPOINT ) );
+        prvTLSConnect( &xNetworkCredentials, &xNetworkContext );
+
+        /* Sends an MQTT Connect packet over the already connected TCP socket,
+         * and waits for connection acknowledgment (CONNACK) packet. */
+        LogInfo( ( "Creating an MQTT connection to %s.\r\n", democonfigAWS_IOT_ENDPOINT ) );
+        prvCreateMQTTConnectionWithBroker( &xMQTTContext, &xNetworkContext );
+
+        /**************************** Subscribe. ******************************/
+
+        /* The client is now connected to the broker. Subscribe to the topic
+         * as specified in mqttexampleTOPIC at the top of this file by sending a
+         * subscribe packet then waiting for a subscribe acknowledgment (SUBACK).
+         * This client will then publish to the same topic it subscribed to, so it
+         * will expect all the messages it sends to the broker to be sent back to it
+         * from the broker. This demo uses QoS1 in Subscribe, therefore, the Publish
+         * messages received from the broker will have QoS1. */
+        LogInfo( ( "Attempt to subscribe to the MQTT topic %s.\r\n", mqttexampleTOPIC ) );
+        prvMQTTSubscribeToTopic( &xMQTTContext );
+
+        /* Process incoming packet from the broker. After sending the subscribe, the
+         * client may receive a publish before it receives a subscribe ack. Therefore,
+         * call generic incoming packet processing function. Since this demo is
+         * subscribing to the topic to which no one is publishing, probability of
+         * receiving Publish message before subscribe ack is zero; but application
+         * must be ready to receive any packet.  This demo uses the generic packet
+         * processing function everywhere to highlight this fact. */
+        xMQTTStatus = MQTT_ProcessLoop( &xMQTTContext, mqttexamplePROCESS_LOOP_TIMEOUT_MS );
+        configASSERT( xMQTTStatus == MQTTSuccess );
+
+        /**************************** Publish and Keep Alive Loop. ******************************/
+        /* Publish messages with QoS1, send and process Keep alive messages. */
+        for( ulPublishCount = 0; ulPublishCount < ulMaxPublishCount; ulPublishCount++ )
+        {
+            LogInfo( ( "Publish to the MQTT topic %s.\r\n", mqttexampleTOPIC ) );
+            prvMQTTPublishToTopic( &xMQTTContext );
+
+            /* Process incoming publish echo, since application subscribed to the same
+             * topic the broker will send publish message back to the application. */
+            LogInfo( ( "Attempt to receive publish message from broker.\r\n" ) );
+            xMQTTStatus = MQTT_ProcessLoop( &xMQTTContext, mqttexamplePROCESS_LOOP_TIMEOUT_MS );
+            configASSERT( xMQTTStatus == MQTTSuccess );
+
+            /* Leave Connection Idle for some time. */
+            LogInfo( ( "Keeping Connection Idle...\r\n\r\n" ) );
+            vTaskDelay( mqttexampleDELAY_BETWEEN_PUBLISHES );
+        }
+
+        /************************ Unsubscribe from the topic. **************************/
+        LogInfo( ( "Unsubscribe from the MQTT topic %s.\r\n", mqttexampleTOPIC ) );
+        prvMQTTUnsubscribeFromTopic( &xMQTTContext );
+
+        /* Process Incoming packet from the broker. */
+        xMQTTStatus = MQTT_ProcessLoop( &xMQTTContext, mqttexamplePROCESS_LOOP_TIMEOUT_MS );
+        configASSERT( xMQTTStatus == MQTTSuccess );
+
+        /**************************** Disconnect. ******************************/
+
+        /* Send an MQTT Disconnect packet over the already connected TCP socket.
+         * There is no corresponding response for the disconnect packet. After sending
+         * disconnect, client must close the network connection. */
+        LogInfo( ( "Disconnecting the MQTT connection with %s.\r\n", democonfigAWS_IOT_ENDPOINT ) );
+        MQTT_Disconnect( &xMQTTContext );
+
+        /* Close the network connection.  */
+        Mbedtls_FreeRTOS_Disconnect( &xNetworkContext );
+
+        /* Wait for some time between two iterations to ensure that we do not
+         * bombard the public test mosquitto broker. */
+        LogInfo( ( "prvMQTTDemoTask() completed an iteration successfully. Total free heap is %u.\r\n", xPortGetFreeHeapSize() ) );
+        LogInfo( ( "Demo completed successfully.\r\n" ) );
+        LogInfo( ( "Short delay before starting the next iteration.... \r\n\r\n" ) );
+        vTaskDelay( mqttexampleDELAY_BETWEEN_DEMO_ITERATIONS );
+    }
+}
+/*-----------------------------------------------------------*/
+
+static void prvTLSConnect( NetworkCredentials_t * pxNetworkCredentials,
+                           NetworkContext_t * pxNetworkContext )
+{
+    BaseType_t xNetworkStatus;
+
+    /* Set the credentials for establishing a TLS connection. */
+    pxNetworkCredentials->pRootCa = democonfigAWS_ROOT_CA_PEM;
+    pxNetworkCredentials->rootCaSize = sizeof( democonfigAWS_ROOT_CA_PEM );
+    pxNetworkCredentials->pClientCert = democonfigCLIENT_CERTIFICATE_PEM;
+    pxNetworkCredentials->clientCertSize = sizeof( democonfigCLIENT_CERTIFICATE_PEM );
+    pxNetworkCredentials->pPrivateKey = democonfigCLIENT_PRIVATE_KEY_PEM;
+    pxNetworkCredentials->privateKeySize = sizeof( democonfigCLIENT_PRIVATE_KEY_PEM );
+
+    /* Attempt to create a mutually authenticated TLS connection. */
+    xNetworkStatus = Mbedtls_FreeRTOS_Connect( pxNetworkContext,
+                                               democonfigAWS_IOT_ENDPOINT,
+                                               democonfigMQTT_BROKER_PORT,
+                                               pxNetworkCredentials,
+                                               TRANSPORT_SEND_RECV_TIMEOUT_MS,
+                                               TRANSPORT_SEND_RECV_TIMEOUT_MS );
+    configASSERT( xNetworkStatus == MBEDTLS_SUCCESS );
+    LogInfo( ( "A mutually authenticated TLS connection established with %s.\r\n",
+               democonfigAWS_IOT_ENDPOINT ) );
+}
+/*-----------------------------------------------------------*/
+static void prvCreateMQTTConnectionWithBroker( MQTTContext_t * pxMQTTContext,
+                                               NetworkContext_t * pxNetworkContext )
+{
+    MQTTStatus_t xResult;
+    MQTTConnectInfo_t xConnectInfo;
+    bool xSessionPresent;
+    TransportInterface_t xTransport;
+    MQTTApplicationCallbacks_t xCallbacks;
+
+    /***
+     * For readability, error handling in this function is restricted to the use of
+     * asserts().
+     ***/
+
+    /* Fill in Transport Interface send and receive function pointers. */
+    xTransport.pNetworkContext = pxNetworkContext;
+    xTransport.send = Mbedtls_FreeRTOS_send;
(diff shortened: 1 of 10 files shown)


FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * http://www.FreeRTOS.org
+ * http://aws.amazon.com/freertos
+ *
+ * 1 tab == 4 spaces!
```
aggarw13, line 25: Is this applicable anymore considering that we are not using tabs anymore for formatting?

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+ */
+
+/*
```
aggarw13, line 28: Should there be `@file` doxygen tags for demo files?

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+ * Demo for showing use of the managed MQTT API using a mutually
+ * authenticated connection.
```
aggarw13, line 30: (Nit)
```suggestion
 * Demo for showing use of the managed MQTT API using a mutually
 * authenticated network connection.
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+ *
+ * The Example shown below uses managed MQTT APIs to create MQTT messages and
+ * send them over the mutually authenticated connection established to the
```
aggarw13, line 33: (Nit)
```suggestion
 * send them over the mutually authenticated network connection established with the
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+ * AWS IoT MQTT broker. This example is single threaded and uses statically
+ * allocated memory. It uses QoS1 for outgoing publishes and to request QoS
+ * level of the incoming publishes.
```
aggarw13, line 36: (Nit) 
```suggestion
 * allocated memory. It uses QoS1 for sending to and receiving messages from the broker.
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/demo_config.h
```diff
@@ -0,0 +1,152 @@
...
+#include "logging_stack.h"
+
+/************ End of logging configuration ****************/
+
+/**
+ * @brief The MQTT client identifier used in this example.  Each client identifier
+ * must be unique so edit as required to ensure no two clients connecting to the
+ * same broker use the same client identifier.
+ *
+ * This is the "Thing Name" in AWS IoT.
```
aggarw13, line 61: Is this a valid statement? Thing Name as the client ID is not strictly required, is it?

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/demo_config.h
```diff
@@ -0,0 +1,152 @@
...
+    #define LIBRARY_LOG_LEVEL    LOG_INFO
+#endif
+#include "logging_stack.h"
+
+/************ End of logging configuration ****************/
+
+/**
+ * @brief The MQTT client identifier used in this example.  Each client identifier
+ * must be unique so edit as required to ensure no two clients connecting to the
+ * same broker use the same client identifier.
```
aggarw13, line 59: Being a grammar nazi!
```suggestion
 * @brief The MQTT client identifier used in this example.  Each client identifier
 * must be unique; so edit as required to ensure that no two clients connecting to
 * the same broker use the same client identifier.
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/demo_config.h
```diff
@@ -0,0 +1,152 @@
...
+ * #define democonfigCLIENT_IDENTIFIER    "insert here."
+ */
+
+/**
+ * @brief Details of the MQTT broker to connect to.
+ *
+ * This is the Thing's Rest API Endpoint for AWS IoT.
+ *
+ * @note Your AWS IoT Core endpoint can be found in the AWS IoT console under
+ * Settings/Custom Endpoint, or using the describe-endpoint API.
```
aggarw13, line 72: For better clarity
```suggestion
 * Settings/Custom Endpoint, or using the describe-endpoint REST API with
 * AWS CLI command line tool.
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/demo_config.h
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * This is the "Thing Name" in AWS IoT.
+ *
+ * #define democonfigCLIENT_IDENTIFIER    "insert here."
+ */
+
+/**
+ * @brief Details of the MQTT broker to connect to.
+ *
+ * This is the Thing's Rest API Endpoint for AWS IoT.
```
aggarw13, line 69: Should probably be in caps
```suggestion
 * This is the Thing's REST API Endpoint for AWS IoT.
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/demo_config.h
```diff
@@ -0,0 +1,152 @@
...
+ * #define democonfigCLIENT_PRIVATE_KEY_PEM    "...insert here..."
+ */
+
+/**
+ * @brief Set the stack size of the main demo task.
+ *
+ * In the Windows port, this stack only holds a structure. The actual
+ * stack is created by an operating system thread.
+ */
+#define democonfigDEMO_STACKSIZE    configMINIMAL_STACK_SIZE
```
aggarw13, line 145: This brings a dependency on the `FreeRTOSConfig.h` file unfortunately where the `configMINIMAL_STACK_SIZE` is defined which would require adding the header include of the `FreeRTOSConfig.h` file in the file....

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/demo_config.h
```diff
@@ -0,0 +1,152 @@
...
+ * "...base64 data...\n"\
+ * "-----END RSA PRIVATE KEY-----\n"
+ *
+ * #define democonfigCLIENT_PRIVATE_KEY_PEM    "...insert here..."
+ */
+
+/**
+ * @brief Set the stack size of the main demo task.
+ *
+ * In the Windows port, this stack only holds a structure. The actual
```
aggarw13, line 142: This comment is not clear to me.. 
Which structure is held onto by the Windows port? 
How does this value affect the ability to create the demo task differently than Windows if it were running on another HW platform?

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/FreeRTOSConfig.h
```diff
@@ -0,0 +1,216 @@
...
+ * the demo which real port should be used to create the virtual port.  The ports
+ * available are displayed on the console when the application is executed.  For
+ * example, on my development laptop setting configNETWORK_INTERFACE_TO_USE to 4
+ * results in the wired network being used, while setting
+ * configNETWORK_INTERFACE_TO_USE to 2 results in the wireless network being
+ * used. */
+#define configNETWORK_INTERFACE_TO_USE            1L
+
+/* The address to which logging is sent should UDP logging be enabled. */
+#define configUDP_LOGGING_ADDR0                   192
```
aggarw13, line 143: Are these macros valid for the demo..? It could be confusing about the presence of this along with the logging stack implementation that re-directs the value to `vLoggingPrintf`...

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+#endif
+#ifndef democonfigCLIENT_PRIVATE_KEY_PEM
+    #error "Please define client private key(democonfigCLIENT_PRIVATE_KEY_PEM) in demo_config.h."
+#endif
+
+/*-----------------------------------------------------------*/
+
+/* Default values for configs. */
+#ifndef democonfigCLIENT_IDENTIFIER
+
```
aggarw13, line 85: Can this new line be avoided? It makes the `#ifndef` guard block less obvious. If uncrustify is adding the new line, may be, the comment block can be moved above the `#ifndef`

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+ * @note Appending __TIME__ to the client id string will reduce the possibility of a
+ * client id collision in the broker. Note that the appended time is the compilation
+ * time. This client id can cause collision, if more than one instance of the same
+ * binary is used at the same time to connect to the broker.
+ */
+    #define democonfigCLIENT_IDENTIFIER    "testClient"__TIME__
+#endif
+
+#ifndef democonfigMQTT_BROKER_PORT
+
```
aggarw13, line 100: Same with the new line here...

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+/* Standard includes. */
+#include <string.h>
+#include <stdio.h>
+
+/* Kernel includes. */
+#include "FreeRTOS.h"
+#include "task.h"
+
+/* FreeRTOS+TCP includes. */
+#include "FreeRTOS_IP.h"
```
aggarw13, line 53: Are these includes required (considering that the TLS implementation of the transport interface is the API that is used in the demo)?

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+static uint16_t usSubscribePacketIdentifier;
+
+/**
+ * @brief Packet Identifier generated when Unsubscribe request was sent to the broker;
+ * it is used to match received Unsubscribe response to the transmitted Unsubscribe
+ * request.
+ */
+static uint16_t usUnsubscribePacketIdentifier;
+
+
```
aggarw13, line 292: Extra new line can be removed
```suggestion
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+/**
+ * @brief The MQTT message published in this example.
+ */
+#define mqttexampleMESSAGE                          "Hello World!"
+
+/**
+ * @brief Dimensions a file scope buffer currently used to send and receive MQTT data
+ * from a socket.
+ */
+#define mqttexampleSHARED_BUFFER_SIZE               ( 500U )
```
aggarw13, line 131: Is this meant to be the same as `NETWORK_BUFFER_SIZE` defined in the `demo_config.h` file?

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+#include "FreeRTOS_IP.h"
+#include "FreeRTOS_Sockets.h"
+
+/* Demo Specific configs. */
+#include "demo_config.h"
+
+/* MQTT library includes. */
+#include "mqtt.h"
+
+/* Transport interface include. */
```
aggarw13, line 62:
```suggestion
/* Transport interface implementation include header for TLS. */
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+/**
+ * @brief Dimensions a file scope buffer currently used to send and receive MQTT data
+ * from a socket.
+ */
+#define mqttexampleSHARED_BUFFER_SIZE               ( 500U )
+
+/**
+ * @brief Time to wait between each cycle of the demo implemented by prvMQTTDemoTask().
+ */
+#define mqttexampleDELAY_BETWEEN_DEMO_ITERATIONS    ( pdMS_TO_TICKS( 5000U ) )
```
aggarw13, line 136: As other time specific macros end with the unit, this one can also carry the unit
```suggestion
#define mqttexampleDELAY_TICKS_BETWEEN_DEMO_ITERATIONS    ( pdMS_TO_TICKS( 5000U ) )
```

aggarw13: Same for the `mqttexampleDELAY_BETWEEN_PUBLISHES`  macro.
Alternatively, these macros can represent only MS values, and the `pdMS_TO_TICKS` can be called before passing the constant to the FreeRTOS function for delay

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+ *
+ * It is the responsibility of the Client to ensure that the interval between
+ * Control Packets being sent does not exceed the this Keep Alive value. In the
+ * absence of sending any other Control Packets, the Client MUST send a
+ * PINGREQ Packet.
+ */
+#define mqttexampleKEEP_ALIVE_TIMEOUT_SECONDS       ( 60U )
+
+/**
+ * @brief Delay between MQTT publishes. Note that the process loop also has a
```
aggarw13, line 154: May be, this makes it clearer?
```suggestion
 * @brief Delay between consecutive cycles of MQTT publish operations in a demo iteration. Note that the process loop also has a
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+ * timeout, so the total time between publishes is the sum of the two delays.
+ */
+#define mqttexampleDELAY_BETWEEN_PUBLISHES          ( pdMS_TO_TICKS( 2000U ) )
+
+/**
+ * @brief Transport timeout in milliseconds for transport send and receive.
+ */
+#define TRANSPORT_SEND_RECV_TIMEOUT_MS              ( 200U )
+
+#define _MILLISECONDS_PER_SECOND                    ( 1000U )                                         /**< @brief Milliseconds per second. */
```
aggarw13, line 164: For consistency with the previous macros in the file, would be good to have the `@brief` comment in the line above the macro.

Same for the `_MILLISECONDS_PER_TICK`

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+/**
+ * @brief The task used to demonstrate the MQTT API.
+ *
+ * @param[in] pvParameters Parameters as passed at the time of task creation. Not
+ * used in this example.
+ */
+static void prvMQTTDemoTask( void * pvParameters );
+
+/**
+ * @brief Sends an MQTT Connect packet over the already connected TCP socket.
```
aggarw13, line 178:
```suggestion
 * @brief Sends an MQTT Connect packet over the already connected TLS over TCP connection.
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+ *
+ * @param pxMQTTContext MQTT context pointer.
+ * @param xNetworkContext network context.
```
aggarw13, line 181: Can you update all the `@param` tags across function declarations to represent the parameter type as input, output or both?
```suggestion
 * @param pxMQTTContext[in, out[ MQTT context pointer.
 * @param xNetworkContext[in, out] network context.
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+/**
+ * @brief Subscribes to the topic as specified in mqttexampleTOPIC at the top of
+ * this file.
+ *
+ * @param pxMQTTContext MQTT context pointer.
+ */
+static void prvMQTTSubscribeToTopic( MQTTContext_t * pxMQTTContext );
+
+/**
+ * @brief  Publishes a message mqttexampleMESSAGE on mqttexampleTOPIC topic.
```
aggarw13, line 195: (Nit)
```suggestion
 * @brief Publishes a message mqttexampleMESSAGE on mqttexampleTOPIC topic.
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+ */
+static void prvMQTTDemoTask( void * pvParameters );
+
+/**
+ * @brief Sends an MQTT Connect packet over the already connected TCP socket.
+ *
+ * @param pxMQTTContext MQTT context pointer.
+ * @param xNetworkContext network context.
+ */
+static void prvCreateMQTTConnectionWithBroker( MQTTContext_t * pxMQTTContext,
```
aggarw13, line 183: (Out of scope of the PR) Is bringing the reconnect logic utility to this repository in the plans?
@abhidixi11

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+ * it is used to match received Unsubscribe response to the transmitted Unsubscribe
+ * request.
+ */
+static uint16_t usUnsubscribePacketIdentifier;
+
+
+/** @brief Static buffer used to hold MQTT messages being sent and received. */
+static MQTTFixedBuffer_t xBuffer =
+{
+    .pBuffer = ucSharedBuffer,
```
aggarw13, line 296: Is this form of initialization compliant with C89?

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+/**
+ * @brief The timer query function provided to the MQTT context.
+ *
+ * @return Time in milliseconds.
+ */
+static uint32_t prvGetTimeMs( void );
+
+/**
+ * @brief Process a response or ack to an MQTT request (PING, SUBSCRIBE
+ * or UNSUBSCRIBE). This function processes PINGRESP, SUBACK, UNSUBACK
```
aggarw13, line 218: It handles PUBACKs as well...

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+ *
+ * @param pxIncomingPacket is a pointer to structure containing deserialized
+ * MQTT response.
+ * @param usPacketId is the packet identifier from the ack received.
+ */
+static void prvMQTTProcessResponse( MQTTPacketInfo_t * pxIncomingPacket,
```
aggarw13, line 224: Would be more appropriate (imo) to call it `prvHandleIncomingAcksOrPingResponse` as it handles all acks as well PING response. 
Is there an advantage of keeping `MQTT` in the name of most of the private functions? I would be fine removing them, especially for this handler which is application-specific. (The event callback is named as `prvEventCallback` without the use of `MQTT` ...)

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+ * @brief Process incoming Publish message.
+ *
+ * @param pxPublishInfo is a pointer to structure containing deserialized
+ * Publish message.
+ */
+static void prvMQTTProcessIncomingPublish( MQTTPublishInfo_t * pxPublishInfo );
+
+/**
+ * @brief The application callback function for getting the incoming publish
+ * and incoming acks reported from the MQTT library.
```
aggarw13, line 237: Also, for receiving Ping responses..

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+ *
+ * @param pxMQTTContext MQTT context pointer.
+ * @param pxPacketInfo Packet Info pointer for the incoming packet.
+ * @param usPacketIdentifier Packet identifier of the incoming packet.
+ * @param pxPublishInfo Deserialized publish info pointer for the incoming
+ * packet.
```
aggarw13, line 243:
```suggestion
 * @param pxPublishInfo Deserialized publish info for the incoming packet if
 * there is an incoming PUBLISH; NULL otherwise
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+static MQTTFixedBuffer_t xBuffer =
+{
+    .pBuffer = ucSharedBuffer,
+    .size    = mqttexampleSHARED_BUFFER_SIZE
+};
+
+/*-----------------------------------------------------------*/
+
+/*
+ * @brief Create the task that demonstrates the Plain text MQTT API Demo.
```
aggarw13, line 303:
```suggestion
 * @brief Create the task that demonstrates the MQTT API Demo over a 
 * mutually authenticated network connection with AWS IoT broker.
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+                 NULL,                     /* Task parameter - not used in this case. */
+                 tskIDLE_PRIORITY,         /* Task priority, must be between 0 and configMAX_PRIORITIES - 1. */
+                 NULL );                   /* Used to pass out a handle to the created task - not used in this case. */
+}
+/*-----------------------------------------------------------*/
+
+static void prvMQTTDemoTask( void * pvParameters )
+{
+    uint32_t ulPublishCount = 0U;
+    const uint32_t ulMaxPublishCount = 5UL;
```
aggarw13, line 322: How about representing it with a `# define` macro instead of stack-scoped constant?

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+    NetworkContext_t xNetworkContext = { 0 };
+    NetworkCredentials_t xNetworkCredentials = { 0 };
+    MQTTContext_t xMQTTContext;
```
aggarw13, line 325: Explicit initialization of the context as well like lines 324-325?

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+    ulGlobalEntryTimeMs = prvGetTimeMs();
+
+    for( ; ; )
+    {
+        /****************************** Connect. ******************************/
+
+        /* Establish a TLS connection with the MQTT broker. This example connects to
+         * the MQTT broker as specified in democonfigAWS_IOT_ENDPOINT and
+         * democonfigMQTT_BROKER_PORT at the top of this file. */
+        LogInfo( ( "Creating a TLS connection to %s.\r\n", democonfigAWS_IOT_ENDPOINT ) );
```
aggarw13, line 340: The port can be logged as well

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+    ( void ) pvParameters;
+
+    ulGlobalEntryTimeMs = prvGetTimeMs();
+
+    for( ; ; )
+    {
+        /****************************** Connect. ******************************/
+
+        /* Establish a TLS connection with the MQTT broker. This example connects to
+         * the MQTT broker as specified in democonfigAWS_IOT_ENDPOINT and
```
aggarw13, line 338: `democonfigAWS_IOT_ENDPOINT` is not specified at the top of the file, but only in the config file...

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+         * democonfigMQTT_BROKER_PORT at the top of this file. */
+        LogInfo( ( "Creating a TLS connection to %s.\r\n", democonfigAWS_IOT_ENDPOINT ) );
+        prvTLSConnect( &xNetworkCredentials, &xNetworkContext );
+
+        /* Sends an MQTT Connect packet over the already connected TCP socket,
```
aggarw13, line 343:
```suggestion
        /* Sends an MQTT Connect packet over the already established TLS connection,
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+
+    for( ; ; )
+    {
+        /****************************** Connect. ******************************/
+
+        /* Establish a TLS connection with the MQTT broker. This example connects to
+         * the MQTT broker as specified in democonfigAWS_IOT_ENDPOINT and
+         * democonfigMQTT_BROKER_PORT at the top of this file. */
+        LogInfo( ( "Creating a TLS connection to %s.\r\n", democonfigAWS_IOT_ENDPOINT ) );
+        prvTLSConnect( &xNetworkCredentials, &xNetworkContext );
```
aggarw13, line 341: Are there no return values to handle failures in the demo app for both TLS and MQTT connect?

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+        /* Sends an MQTT Connect packet over the already connected TCP socket,
+         * and waits for connection acknowledgment (CONNACK) packet. */
+        LogInfo( ( "Creating an MQTT connection to %s.\r\n", democonfigAWS_IOT_ENDPOINT ) );
+        prvCreateMQTTConnectionWithBroker( &xMQTTContext, &xNetworkContext );
+
+        /**************************** Subscribe. ******************************/
+
+        /* The client is now connected to the broker. Subscribe to the topic
+         * as specified in mqttexampleTOPIC at the top of this file by sending a
+         * subscribe packet then waiting for a subscribe acknowledgment (SUBACK).
```
aggarw13, line 352: The "then waiting for a subscribe acknowlegement (SUBACK)" phrase can be misleading about the `prvMQTTSubscribeToTopic` being a blocking call on the arrival of `SUBACK`. 
I think it should be clearly mentioned that waiting for `SUBACK` is only done in the `ProcessLoop` call

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+
+        /**************************** Publish and Keep Alive Loop. ******************************/
+        /* Publish messages with QoS1, send and process Keep alive messages. */
+        for( ulPublishCount = 0; ulPublishCount < ulMaxPublishCount; ulPublishCount++ )
+        {
+            LogInfo( ( "Publish to the MQTT topic %s.\r\n", mqttexampleTOPIC ) );
+            prvMQTTPublishToTopic( &xMQTTContext );
+
+            /* Process incoming publish echo, since application subscribed to the same
+             * topic the broker will send publish message back to the application. */
```
aggarw13, line 378: (Nit)
```suggestion
             * topic, the broker will send publish message back to the application. */
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+            /* Leave Connection Idle for some time. */
+            LogInfo( ( "Keeping Connection Idle...\r\n\r\n" ) );
+            vTaskDelay( mqttexampleDELAY_BETWEEN_PUBLISHES );
+        }
+
+        /************************ Unsubscribe from the topic. **************************/
+        LogInfo( ( "Unsubscribe from the MQTT topic %s.\r\n", mqttexampleTOPIC ) );
+        prvMQTTUnsubscribeFromTopic( &xMQTTContext );
+
+        /* Process Incoming packet from the broker. */
```
aggarw13, line 392:
```suggestion
        /* Process incoming UNSUBACK packet from the broker. */
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+        xMQTTStatus = MQTT_ProcessLoop( &xMQTTContext, mqttexamplePROCESS_LOOP_TIMEOUT_MS );
+        configASSERT( xMQTTStatus == MQTTSuccess );
+
+        /**************************** Disconnect. ******************************/
+
+        /* Send an MQTT Disconnect packet over the already connected TCP socket.
```
aggarw13, line 398:
```suggestion
        /* Send an MQTT Disconnect packet over the already connected TLS over TCP connection.
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+        LogInfo( ( "Creating an MQTT connection to %s.\r\n", democonfigAWS_IOT_ENDPOINT ) );
+        prvCreateMQTTConnectionWithBroker( &xMQTTContext, &xNetworkContext );
+
+        /**************************** Subscribe. ******************************/
+
+        /* The client is now connected to the broker. Subscribe to the topic
+         * as specified in mqttexampleTOPIC at the top of this file by sending a
+         * subscribe packet then waiting for a subscribe acknowledgment (SUBACK).
+         * This client will then publish to the same topic it subscribed to, so it
+         * will expect all the messages it sends to the broker to be sent back to it
```
aggarw13, line 354: I think a general overview of the functionality in the demo task can be provided at the start of this function, which will provide a good context for these operations in each iteration. 
This comment seems to provide some of that overview, but it could be helpful to provide the complete overview earlier, and instead keep this comment specific to just the subscribe operation

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+
+    /* Send MQTT CONNECT packet to broker. LWT is not used in this demo, so it
+     * is passed as NULL. */
+    xResult = MQTT_Connect( pxMQTTContext,
+                            &xConnectInfo,
+                            NULL,
+                            mqttexampleCONNACK_RECV_TIMEOUT_MS,
+                            &xSessionPresent );
+
+    if( xResult != MQTTSuccess )
```
aggarw13, line 501: Most of the other error code handling is done as assert in the file....Can you make this consistent? (I personally prefer that the error codes be handled and logged instead of `configASSERT`, however, that would entail a decent amount of code addition; thus, I would be fine with using asserts for this PR)

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+
+    /* Get the current tick count. */
+    xTickCount = xTaskGetTickCount();
+
+    /* Convert the ticks to milliseconds. */
+    ulTimeMs = ( uint32_t ) xTickCount * _MILLISECONDS_PER_TICK;
+
+    /* Reduce ulGlobalEntryTimeMs from obtained time so as to always return the
+     * elapsed time in the application. */
+    ulTimeMs = ( uint32_t ) ( ulTimeMs - ulGlobalEntryTimeMs );
```
aggarw13, line 696: How about when the `xTaskGetTickCount` value over flows? In that case, this subtraction would overflow as well..

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,701 @@
...
+
+    /* The client identifier is used to uniquely identify this MQTT client to
+     * the MQTT broker. In a production device the identifier can be something
+     * unique, such as a device serial number. */
+    xConnectInfo.pClientIdentifier = democonfigCLIENT_IDENTIFIER;
+    xConnectInfo.clientIdentifierLength = ( uint16_t ) strlen( democonfigCLIENT_IDENTIFIER );
+
+    /* Set MQTT keep-alive period. It is the responsibility of the application to ensure
+     * that the interval between Control Packets being sent does not exceed the Keep Alive value.
+     * In the absence of sending any other Control Packets, the Client MUST send a PINGREQ Packet. */
```
aggarw13, line 490: The comment about it being the "responsibility of the application to ensure that the interval between Control packets being sent does not exceed Keep Alive" can be misleading about an interval larger than the keep-alive being an incorrect use by the application. 
The statement can be updated to simply say:
```suggestion
    /* Set MQTT keep-alive period. If the application does not send packets at an interval less than
     * the keep-alive period, the MQTT library will send PINGREQ packets. */ 
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
abhidixi11: Yes, we will have to, it will happen separately after this demo.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
leegeth: Yes correct. We will probably have to remove it from all the files in this repo. I will remove it from the files in this PR. Thanks

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
leegeth: I don't think it is a must.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
leegeth: Yea. This may not be required. Thanks

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
leegeth: I have observed the addition of a line in here. I have thought about moving the comments above the #ifdef check as well. However, I thought that the comments will be better if those are right above the #define. I think the indentation does the job for making the #ifndef guard obvious. So I think it is better to leave as is.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
leegeth: Same reply as in the previous comment.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
leegeth: Yea. Good catch. Will fix this.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
leegeth: Agree.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
leegeth: Yea there is a task tracking it.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
leegeth: I think the name is good enough as it is processing the response from the broker. The documentation also talks about the specific responses it is handling. 

Also about the MQTT in names, I think it is ok to keep those as names as these functions are about MQTT related functionalities in this demo.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
leegeth: No. Good catch. Will update it.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
leegeth: We are using configAsserts in the demo and hence no return is required.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
leegeth: Agree.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/demo_config.h
leegeth: yes Correct. Will update the statement.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/demo_config.h
leegeth: Yes nice catch. Will add the header.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/demo_config.h
leegeth: This is how FreeRTOS is ported on Windows. Please find the details about that here https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/6199b72fbf57a7c5b3d7b195a3bd1446779314cd/portable/MSVC-MingW/port.c#L88

For other platforms port layer helps with this

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/FreeRTOSConfig.h
leegeth: It is used by vLoggingPrintf. So it is needed.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
leegeth: I don't have strong opinion about either the macro or a stack-scoped constant. Keeping it in the stack scope is just fine as the usage is limited to the scope of prvMQTTDemoTask function. I don't see any major advantage in keeping it as a macro.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
leegeth: Overflow is handled by ( uint32_t ) ( ulTimeMs - ulGlobalEntryTimeMs ).

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -107,60 +102,66 @@
...
  * PINGREQ Packet.
  */
-#define mqttexampleKEEP_ALIVE_TIMEOUT_SECONDS       ( 60U )
+#define mqttexampleKEEP_ALIVE_TIMEOUT_SECONDS             ( 60U )
 
 /**
- * @brief Delay between MQTT publishes. Note that the process loop also has a
- * timeout, so the total time between publishes is the sum of the two delays.
+ * @brief Delay between consecutive cycles of MQTT publish operations in a
+ * demo iteration in ticks.
```
aggarw13, line 144: (Nit) Suggested would be easier to read
```suggestion
 * @brief Delay (in ticks) between consecutive cycles of MQTT publish operations in a
 * demo iteration.
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -213,32 +214,33 @@ static uint32_t prvGetTimeMs( void );
 
 /**
  * @brief Process a response or ack to an MQTT request (PING, SUBSCRIBE
- * or UNSUBSCRIBE). This function processes PINGRESP, SUBACK, UNSUBACK
+ * or UNSUBSCRIBE). This function processes PINGRESP, PUBACK, SUBACK, and
```
aggarw13: The list of request packets in parentheses should also mention PUBLISH 🙂

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -213,32 +214,33 @@ static uint32_t prvGetTimeMs( void );
...
- * @param pxPublishInfo is a pointer to structure containing deserialized
+ * @param[in] pxPublishInfo is a pointer to structure containing deserialized
  * Publish message.
  */
 static void prvMQTTProcessIncomingPublish( MQTTPublishInfo_t * pxPublishInfo );
 
 /**
- * @brief The application callback function for getting the incoming publish
- * and incoming acks reported from the MQTT library.
+ * @brief The application callback function for getting the incoming publishes,
```
aggarw13, line 236: (Nit)
```suggestion
 * @brief The application callback function for getting incoming publishes,
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -314,31 +316,48 @@ void vStartSimpleMQTTDemo( void )
...
-    MQTTContext_t xMQTTContext;
+    MQTTContext_t xMQTTContext = { 0 };
     MQTTStatus_t xMQTTStatus;
 
     /* Remove compiler warnings about unused parameters. */
     ( void ) pvParameters;
 
+    /* Store the entry time of the demo application. This entry time is used for
+     * calculating the elapsed time in application, by the time utility function
+     * used by MQTT library. */
```
aggarw13, line 345: The comment is confusing here in the phrases of "calculating the elapse time in application" which could suggest that the application is calculating the elapsed time...
"by the time utility function used by MQTT library" - This can be confusing to the reader about which time utility library is used by MQTT if they don't have the understanding of the library API yet.


```suggestion
    /* Set the entry time of the demo application. This entry time will be used
     * to calculate relative time elapsed in the execution of the demo application,
     * by the timer utility function that is provided to the MQTT library.
     */
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -348,10 +367,12 @@ static void prvMQTTDemoTask( void * pvParameters )
...
          * as specified in mqttexampleTOPIC at the top of this file by sending a
          * subscribe packet then waiting for a subscribe acknowledgment (SUBACK).
-         * This client will then publish to the same topic it subscribed to, so it
-         * will expect all the messages it sends to the broker to be sent back to it
-         * from the broker. This demo uses QoS1 in Subscribe, therefore, the Publish
-         * messages received from the broker will have QoS1. */
+         * The function #prvMQTTSubscribeToTopic will not wait to receive a SUBACK,
+         * but the function #MQTT_ProcessLoop will attempt to receive the SUBACK
+         * from network and if a SUBACK is received, application will be notified
+         * through the callback registered(#prvEventCallback for this application).
```
aggarw13, line 373:
```suggestion
         * through the callback registered ( #prvEventCallback for this application ).
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -456,13 +479,15 @@ static void prvCreateMQTTConnectionWithBroker( MQTTContext_t * pxMQTTContext,
     xTransport.send = TLS_FreeRTOS_send;
     xTransport.recv = TLS_FreeRTOS_recv;
 
-    /* Application callbacks for receiving incoming published and incoming acks
+    /* Application callback for receiving incoming published and incoming acks
      * from MQTT library. */
-    xCallbacks.appCallback = prvEventCallback;
-    xCallbacks.getTime = prvGetTimeMs;
+    xUserCallback = prvEventCallback;
```
aggarw13, line 484: I was suggested by @aggarg to avoid these stack variables but instead just pass the `prEventCallback` and `prvGetTimeMs` arguments directly in the `MQTT_Init` call...

leegeth: Oh ok. Sure. Will keep it consistent.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,711 @@
...
+         * There is no corresponding response for the disconnect packet. After sending
+         * disconnect, client must close the network connection. */
+        LogInfo( ( "Disconnecting the MQTT connection with %s.\r\n", democonfigAWS_IOT_ENDPOINT ) );
+        MQTT_Disconnect( &xMQTTContext );
+
+        /* Close the network connection.  */
+        TLS_FreeRTOS_Disconnect( &xNetworkContext );
+
+        /* Wait for some time between two iterations to ensure that we do not
+         * bombard the public test mosquitto broker. */
```
gkwicker, line 428: I thought this demo was written to use AWS IoT?  Does it use the public test mosquitto broker?  If so, it shouldn't.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,711 @@
...
+}
+/*-----------------------------------------------------------*/
+
+static void prvTLSConnect( NetworkCredentials_t * pxNetworkCredentials,
+                           NetworkContext_t * pxNetworkContext )
+{
+    BaseType_t xNetworkStatus;
+
+    /* Set the credentials for establishing a TLS connection. */
+    pxNetworkCredentials->pRootCa = ( const unsigned char * ) democonfigAWS_ROOT_CA_PEM;
```
gkwicker, line 443: I believe if you just remove the `AWS` from these defines and adjust the comments, this demo won't have anything to do with AWS IoT.  It would be great to provide instructions for using the demo with an instance of mosquitto as well as with AWS IoT.  Mosquitto would be the most important of these given that this is non-AWS code.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,711 @@
...
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * http://www.FreeRTOS.org
+ * http://aws.amazon.com/freertos
+ */
+
+/*
+ * Demo for showing use of the managed MQTT API using a mutually
```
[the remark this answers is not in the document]
leegeth: Thanks. I see your point and agree with that. There are 2 ways in which MQTT library can be used. 
1. Use only the serializer and deserializer APIs.
2. Managed MQTT API - which will make use of the serializer and deserializer APIs internally. These APIs will also be managing the send and receive on the network and state of MQTT messages with QoS > Q0S0.

In here, I think falling back to MQTT API would be good as the serailizers can be referred by serializers and deserializers itself.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,711 @@
...
+ * authenticated network connection.
+ *
+ * The Example shown below uses managed MQTT APIs to create MQTT messages and
+ * send them over the mutually authenticated network connection established with the
+ * AWS IoT MQTT broker. This example is single threaded and uses statically
```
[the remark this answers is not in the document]
leegeth: Agree. Will make the instructions be more generic. Thanks

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,711 @@
...
+#ifndef democonfigCLIENT_IDENTIFIER
+
+/**
+ * @brief The MQTT client identifier used in this example.  Each client identifier
+ * must be unique so edit as required to ensure no two clients connecting to the
+ * same broker use the same client identifier.
+ *
+ * @note Appending __TIME__ to the client id string will reduce the possibility of a
+ * client id collision in the broker. Note that the appended time is the compilation
+ * time. This client id can cause collision, if more than one instance of the same
```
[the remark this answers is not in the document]
leegeth: "Could a random number be used instead of a compile-time constant? "

This is a default value generated only if the a config is not already defined in demo_config.h. It is done as a compile time constant as this path is used only in default case and customer application would be able to configure a unique client id.

I will add the note as you suggested. Thanks

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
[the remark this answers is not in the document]
gkwicker: Just a couple of more "managed" to delete on lines 27 and 320 :-)

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,711 @@
...
+ * Demo for showing use of the managed MQTT API using a mutually
+ * authenticated network connection.
+ *
+ * The Example shown below uses MQTT APIs to create MQTT messages and send them
+ * over the mutually authenticated network connection established with the
+ * MQTT broker. This example is single threaded and uses statically allocated
+ * memory. It uses QoS1 for sending to and receiving messages from the broker.
+ *
+ * A mutually authenticated TLS connection is used to connect to the
+ * MQTT message broker in this example. Define democonfigROOT_CA_PEM,
```
gkwicker, line 36: I think they also need to define `democonfigMQTT_BROKER_ENDPOINT` so let's mention that as well.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_mutual_auth/DemoTasks/MutualAuthMQTTExample.c
```diff
@@ -0,0 +1,711 @@
...
+    /* Fill in Transport Interface send and receive function pointers. */
+    xTransport.pNetworkContext = pxNetworkContext;
+    xTransport.send = TLS_FreeRTOS_send;
+    xTransport.recv = TLS_FreeRTOS_recv;
+
+    /* Initialize MQTT library. */
+    xResult = MQTT_Init( pxMQTTContext, &xTransport, prvGetTimeMs, prvEventCallback, &xBuffer );
+    configASSERT( xResult == MQTTSuccess );
+
+    /* Many fields not used in this demo so start with everything at 0. */
```
gkwicker, line 485: nit: _"Many fields __are__ not used..."_ - same comment at lines 521, 568, 587

leegeth:
Waiting for the build checks to pass before this change can be merged. Build checks are failing due to microchip project dependency, which will be addressed by a different PR.

PR #186, opened 2020-07-31
Update wolfSSL to the latest version(v.4.4.0)

TakayukiMatsuo:
Update wolfSSL to the latest version(v4.4.0)

Description
-----------
This update includes followings:
1. update files and folders under the FreeRTOS-Plus/Source/WolfSSL folder to the latest
2. in the FreeRTOS-Plus/Demo/FreeRTOS_Plus_WolfSSL_Windows_Simulator 
    - ca-cert.pem and server-cert.pem files are updated 
    - user_settings.h file is introduced to provide users customizing cipher algorithms.
    - preprocessor macro "WOLFSSL_USER_SETTINGS" is added to the WIN32.vcxproj so as wolfSSL to refer the user_settings.h

Test Steps
-----------
Build RTOSDemo.exe with the solution file in the FreeRTOS-Plus/Demo/FreeRTOS_Plus_WolfSSL_Windows_Simulator and then run it with Debugger. 

Related Issue
-----------
None


RichardBarry:
Thanks for the update.  Perhaps in future we should pull this code in as a submodule - although after accepting this pull request we would normally remove the files the demos don't build to reduce the size on disk which would not be possible as a submodule....

Our current demo (which is rather old and perhaps should be regenerated! https://github.com/FreeRTOS/FreeRTOS/tree/master/FreeRTOS-Plus/Demo/FreeRTOS_Plus_WolfSSL_Windows_Simulator) generates the warning "For timing resistance / side-channel attack prevention consider using harden options".  I fixed that (in my local copy) by adding -DWC_NO_HARDEN to the compiler options - but don't think that is the intended way.  Are we missing a local configuration file or this a symptom of having generated that demo using a much older WolfSSL version?

TakayukiMatsuo:
Hi Richard,
Thank you for your review. The warnings came because some macros for timing resistance were missing in user_settings.h which is added this time. I confirmed the warning has gone by adding the macros in the file.

PR #43, opened 2020-03-23
Resolve stdio and stream_buffer link errors in RX64M demo (#42)

befree68:
Resolve stdio and stream_buffer link errors in RX64M demo (#42)

Description
-----------
`.cproject`: Added stdio.h to standard library build. This is required because Common_Demo_Source\Minimal\MessageBufferDemo.c (and others) use `sprintf`.

`.project`: Included stream_buffer.c in the FreeRTOS source directory. This is required because Common_Demo_Source\Minimal\MessageBufferDemo.c uses numerous stream buffer functions such as `vStreamBufferDelete`.

Test Steps
-----------
See #42 for reproduction steps.
Tested by doing a git clone of this fork into a clean workspace, importing the project into Renesas E2 Studio IDE, and pressing "Build All". Verified that no errors occurred. Then loaded it onto the development board and verified that LED0 was blinky.

Related Issue
-----------
N/A

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #42, opened 2020-03-23
[BUG] Renesas RSK64F Demo build errors (linker)


befree68:
**Describe the bug**
There are 33 errors when trying to build the RX600_RX64M_RSK_Renesas_e2studio demo with the latest E2 Studio IDE.

**Target**
- Development board: Renesas RSK+RX64M (R5F564MLLCDFC processor)
- Instruction Set Architecture: RXv2
- IDE and version: Renesas E2 Studio Version 3.0.1.09
- Toolchain and version: RX Compiler CC-RX V2.01.00

**Host**
- Host OS: Windows 10
- Version: 10.0.15063 Build 15063

**To Reproduce**
You do not need hardware to reproduce this issue. Simply download the latest E2 Studio IDE from Renesas website [Windows only] and import the RX64M demo as described on the FreeRTOS website (https://www.freertos.org/RX64M_RTOS_Renesas_GCC_e2studio.html). The initial build fails with 33 linker errors related: 2 related to undefined ref to sprint and 31 related to undefined refs to stream buffer symbols.

**Expected behavior**
Building the demo project should complete successfully without any user modifications required.

**Screenshots**
N/A

**Additional context**
Console output:
```
Description	Resource	Path	Location	Type
E0562310: Undefined external symbol "_xStreamBufferReceiveFromISR" referenced in ".\Source\Common_Demo_Source\Minimal\StreamBufferDemo.obj"	RTOSDemo		 	C/C++ Problem
E0562310: Undefined external symbol "_xStreamBufferReceiveFromISR" referenced in ".\Source\Common_Demo_Source\Minimal\MessageBufferDemo.obj"	RTOSDemo		 	C/C++ Problem
E0562310: Undefined external symbol "_xStreamBufferSend" referenced in ".\Source\Common_Demo_Source\Minimal\MessageBufferDemo.obj"	RTOSDemo		 	C/C++ Problem
E0562310: Undefined external symbol "_xStreamBufferSend" referenced in ".\Source\Common_Demo_Source\Minimal\MessageBufferAMP.obj"	RTOSDemo		 	C/C++ Problem
E0562310: Undefined external symbol "_xStreamBufferSend" referenced in ".\Source\Common_Demo_Source\Minimal\StreamBufferDemo.obj"	RTOSDemo		 	C/C++ Problem
[...]

PR #195, opened 2020-08-11
RSK64M Demo - Fix build errors

dachalco:
**Description:**

- Project was not linking for the demo's baseline IDE/toolchain versions. Some minor IDE/toolchain version upgrades were necessary as the older versions, used in baseline, can no longer be downloaded from vendor site. 
- Recent updates to task notifications made them an array. Added newly required definition of `configTASK_NOTIFICATION_ARRAY_ENTRIES`

**Testing:**
Follow this [FreeRTOS Demo Guide](https://www.freertos.org/RX64M_RTOS_Renesas_GCC_e2studio.html?_ga=2.111024261.1472790108.1597090048-1132513474.1578950675) and verify successful build and run demo on board

**Side Note:**
Originally wanted to upgrade project for latest version of the IDE (Demo was baselined for v3.0). However the 'upgrade project' tool did not work for v7.8.0. There's version jumps that cannot occur. I think one way to do this successfully would be iteratively jump between major versions. I also don't have the the board to test on. As the project upgrade was turning into a lengthy endeavor, I decided to defer it from this PR, and focus on core issue of getting build to work. 

**Related Issue:** 
See [forum post](https://forums.freertos.org/t/renesas-rsk64m-demo-build-errors-undefined-external-symbol-sprintf/9286)


FreeRTOS/Demo/RX600_RX64M_RSK_Renesas_e2studio/.cproject
```diff
@@ -151,5 +153,9 @@
 		<autodiscovery enabled="true" problemReportingEnabled="true" selectedProfileId=""/>
 	</storageModule>
 	<storageModule moduleId="org.eclipse.cdt.core.LanguageSettingsProviders"/>
-	<storageModule moduleId="refreshScope"/>
+	<storageModule moduleId="refreshScope" versionNumber="2">
+		<configuration configurationName="HardwareDebug">
+			<resource resourceType="PROJECT" workspacePath="/RTOSDemo"/>
```
lundinc2, line 158: Not super familiar with this IDE, but this looks like it might be something specific to your personal workspace?

FreeRTOS/Demo/RX600_RX64M_RSK_Renesas_e2studio/makefile.init
```diff
@@ -2,7 +2,7 @@
 # Automatically-generated file. Do not edit!
 ################################################################################
 
-export INC_RX=C:\devtools\Renesas\RX\2_1_0\include
-export RXC_LIB=C:\devtools\Renesas\RX\2_1_0\bin
-export BIN_RX=C:\devtools\Renesas\RX\2_1_0\bin
-PATH := $(PATH):C:\devtools\Renesas\RX\2_1_0\bin
\ No newline at end of file
+export INC_RX=C:\PROGRA~2\Renesas\RX\2_8_1\include
```
lundinc2, line 5: This looks kind of mangled, compared to Program Files (x86) which was used above

FreeRTOS/Demo/RX600_RX64M_RSK_Renesas_e2studio/.cproject
dachalco: I'll see if these changes can be omitted

FreeRTOS/Demo/RX600_RX64M_RSK_Renesas_e2studio/makefile.init
dachalco: The file is generated by IDE/toolchain. IMHO this file should not have been added to Git in the first place.
It should get overwritten by user's IDE when they import and try to build

FreeRTOS/Demo/RX600_RX64M_RSK_Renesas_e2studio/.cproject
NoMaY-jp: This change is one of usual behaviors of the e2 studio. I'm not familiar with other Eclipse based IDEs such as NXP's LPCXpresso and ST's SW4STM32 and so on. But Renesas's e2 studio do as follows.

(1) The e2 studio does not output 'perfect all' entries to the .cprojct file.
(2) I.e. only necessary (and something that e2 studio want to output) entries are output to the .cprojct file.
(3) When you change settings of your project, one or combination of the following changes in the .cproject file happen.
(3-1) Only the exact entries are changed.
(3-2) Only the exact entries are added.
(3-3) Sometimes entries which are unchanged but displayed in the same properties page are updated/added.
(3-4) Sometimes entries something like general purpos entries are updated/added.

This 'refreshScope' entry is the case (3-4).

FreeRTOS/Demo/RX600_RX64M_RSK_Renesas_e2studio/makefile.init
NoMaY-jp: In case of the latest e2 studio v7.8.0(32bit version) and v2020-07(64bit version), the following files are not necessary. (These files are no longer created when you create new proect. Only very old versions of the e2 studio created/used these files.)

(1) .HardwareDebuglinker
(2) .info
(3) makefile.init

NoMaY-jp:
Hello,

I have an impression that this and the following solution is not better for this build error bug 
because demo functions for StreamBuffer/MessageBuffer are not perfomed in this RTOSDemo program 
i.e. the main_full() does not call demo functions for StreamBuffer/MessageBuffer.

Renesas RSK64M demo build errors (undefined external symbol _sprintf)
[forums.freertos.org/t/renesas-rsk64m-demo-build-errors-undefined-external-symbol-sprintf/9286](https://forums.freertos.org/t/renesas-rsk64m-demo-build-errors-undefined-external-symbol-sprintf/9286)

[BUG] Renesas RSK64F Demo build errors (linker) #42
[github.com/FreeRTOS/FreeRTOS/issues/42](https://github.com/FreeRTOS/FreeRTOS/issues/42)

Resolve stdio and stream_buffer link errors in RX64M demo (#42) #43
[github.com/FreeRTOS/FreeRTOS/pull/43](https://github.com/FreeRTOS/FreeRTOS/pull/43)

I think that better way is to exclude (at least) the following files from build.
```
MessageBufferAMP.c
MessageBufferDemo.c
StreamBufferDemo.c
StreamBufferInterrupt.c
```
Probably more files are not used in this RTOSdemo program because main_full() calls only the following functions.
```
void main_full( void )
{
    /* Start all the other standard demo/test tasks.  They have no particular
    functionality, but do demonstrate how to use the FreeRTOS API and test the
    kernel port. */
    vStartInterruptQueueTasks();
    vStartDynamicPriorityTasks();
    vStartBlockingQueueTasks( mainBLOCK_Q_PRIORITY );
    vCreateBlockTimeTasks();
    vStartCountingSemaphoreTasks();
    vStartGenericQueueTasks( tskIDLE_PRIORITY );
    vStartRecursiveMutexTasks();
    vStartSemaphoreTasks( mainSEM_TEST_PRIORITY );
    vStartMathTasks( mainFLOP_TASK_PRIORITY );
    vStartTimerDemoTask( mainTIMER_TEST_PERIOD );
    vStartQueueOverwriteTask( mainQUEUE_OVERWRITE_PRIORITY );
    vStartEventGroupTasks();
    vStartLEDFlashTasks( mainFLASH_PRIORITY );

   ...omit...
```
Best regards,
NoMaY

NoMaY-jp:
Hello @dachalco,

I have two questions.

(1) What is the change of QueueGenericCreateStatic_harness.c? Where it comes from?
(2) If FreeRTOSConfig.h is not changed, what happens? (Because I wonder that necessity of this definition seems that the backward compatibility is lost...)

Best regards,
NoMaY

NoMaY-jp:
Hello,

Now I understand everything perfectly. We have to exclude MessageBuffer*.c and StreamBuffer*.c using resource filter like the following screen copy of the e2 studio's properties page. Some files (alt*.c, com*.c, cr*.c, sp_flop.c) had been already excluded but newly added files (i.e. MessageBuffer*.c and StreamBuffer*.c) has not been excluded because the project has not been updated after the project was created. And maybe TaskNotifyArray.c has to be excluded.

A screen copy of the e2 studio's properties page is here.
[https://forums.freertos.org/t/renesas-rsk64m-demo-build-errors-undefined-external-symbol-sprintf/9286/5](https://forums.freertos.org/t/renesas-rsk64m-demo-build-errors-undefined-external-symbol-sprintf/9286/5)

Best regards,
NoMaY

NoMaY-jp:
Hello @dachalco,

Which version of e2 studio do you use? Because your .project has no longer supported following nature  therefore it seems that you are using very old version of e2 studio.
```
        <nature>com.renesas.cdt.core.kpitcnature</nature>
        <nature>com.renesas.cdt.core.kpitccnature</nature>
```

Best regards,
NoMaY

dachalco:
Hello @NoMaY-jp 

Thank you for pointing out the existing PR #43. I'll aim to get  PR #43  as it's older and from the community! 
Due to current timelines, we want to include a fix that has been verified/tested ASAP. I like your suggestions and we would be welcome to a subsequent enhancement PR of this demo project that might exclude unnecessary files as you pointed out.



> (1) What is the change of QueueGenericCreateStatic_harness.c? Where it comes from?

This was an unrelated stale/reverted commit that I should have deleted from the git history, but it accidentally made into my PR. It has no changes in this PR.

> (2) If FreeRTOSConfig.h is not changed, what happens? (Because I wonder that necessity of this definition seems that the backward compatibility is lost...)

Task notifications were recently upgraded to an array of entries, instead of being a single entry. Backward compatibiltiy is retained as the legacy `xTaskNotify*` APIs still exist and now work by always using the `0th` notification entry, of the total `configTASK_NOTIFICATION_ARRAY_ENTRIES` which has a default array length of `1` as defined in [FreeRTOS.h#852](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/master/include/FreeRTOS.h#L852)

Specifically for this demo however, per the FreeRTOS porting guide this demo implements a satisfactory subset of `./FreeRTOS/Demo/Common/Minimal`. With the addition of notification arrays, comes the addition of new tests for `xTaskNotify*Indexed`. Those tests require at least 3 notification entries which is why I added 
```c
#define configTASK_NOTIFICATION_ARRAY_ENTRIES  ( 3 )
```
While those tests are compiled but not linked in, as you found with some other functions, they may be desired within `main_full` in the future and enable us to quickly add the call in `main_full` for any local testing. 

As for IDE versions content, I added a 'Side Note' to this PR description which provides more details and explanation on the decision. In short, upgrading project for newer versions can be a separate PR.

NoMaY-jp:
Hello @dachalco,

Thank you for your reply. Accept/not accept will be decided by Amazon (I'm not sure whether you are Amazon's engineer or not) but I have an impression that excluding MessageBuffer*.c, StreamBuffer*.c and TaskNotifyArray.c is more 'natural and quick and certain' solution especially for your 'ASAP' requirement after our understanding the fact of 'compiled but not linked in', isn't it?

Regarding to IDE version, I understand your 'ASAP' requirement and your plan of 'Side Note'. (But accept/not accept will be decided by Amazon.)

Best regards,
NoMaY

RichardBarry:
Not read this whole thread so may miss the point - but in managed make Eclipse projects it is normal to use resource filters or virtual links to ensure only the files used by a project are built with the project.  There is some information here: https://www.freertos.org/Project_Workspace_Relative_File_Paths_Eclipse.html although the method described on that page is no longer the preferred method it does still work.

NoMaY-jp:
Hello @dachalco,

I get another idea to build the RX64M Demo project. Please do the following steps.

(1) Remove MessageBuffer*.c, StreamBuffer*.c and TaskNotifyArray.c using Windows explorer.
(2) Import the RX64M Demo project on your e2 studio.
(3) Build the RX64M Demo project on your e2 studio.

I think that you are able to build the RX64M Demo project because removed files are not used.

And regarding to the added 'Side Note', I was able to build the RX64M Demo project by the following steps using e2 studio v7.8.0 and FreeRTOS zip file v10.3.1.

(11) Import the project
(12) Upgrade the project
(13) Select toolchain type and toolchain version in the project setting properties page.
(14) Exclude MessageBuffer*.c and StreamBuffer*.c from build (Using 'exclude from build' was my tentative method)
(15) Then build.

Probably the latest FreeRTOS repository needs excluding TaskNotifyArray.c from build.

I think that you forget doing step (13).

Best regards,
NoMaY

NoMaY-jp:
Hello @dachalco,

I posted screen copies which show the steps to build RSK64M demo using e2 studio v7.8.0 and FreeRTOS v10.3.1.

[https://forums.freertos.org/t/renesas-rsk64m-demo-build-errors-undefined-external-symbol-sprintf/9286/6](https://forums.freertos.org/t/renesas-rsk64m-demo-build-errors-undefined-external-symbol-sprintf/9286/6)

Best regards,
NoMaY

PR #197, opened 2020-08-11
Remove unused wolfSSL files.

mingyue86010:
Description
-----------
This PR remove the files in wolfSSL lib that are not used by demo project.

Test Steps
-----------

Related Issue
-----------


mingyue86010:
Currently the demos let the wolfSSL lib to choose the best available cipher. The wolfSSL user guide only say the wolfSSL would choose the "most secure" cipher without further detailed information. If we can explictly define the cipher suit and remove those unused ciphers, the size of wolfSSL lib can be further greatly reduced.

PR #207, opened 2020-08-20
added Yuhui's basic RISC-V basic PMP demo

n9wxu:
Created a demo for the RISC-V PMP.

Description
-----------
This demo is more of a lab to demonstrate configuring the linker script and PMP initialization code to lock the data memory and prevent execution.


RichardBarry:
Pleased to see this.  Are there any files we can remove from the freedom metal directory to keep the file count down?  This is an equivalent, although it too seems to have a lot of files: https://github.com/FreeRTOS/FreeRTOS/tree/master/FreeRTOS/Demo/RISC-V_RV32_SiFive_HiFive1-RevB_FreedomStudio/freedom-metal

n9wxu:
Since this demo is barely the blinky demo I am sure we can strip quite a few of those files away.

yuhui-zheng:
Please don't put my name on this, this is not what I intended to put out. More in email.

PR #212, opened 2020-08-25
FreeRTOS+TCP /multi /IPv6: FreeRTOS_TCP_WIN

htibosch:
Description
-----------
The purpose of this PR is to get the sources reviewed. Please do not used them for any project.

FreeRTOS_WIN.c needs no changed for either /multi or /IPv6. I'm adding it for completeness only.
I only removed an #include which was not necessary:

~~~c
#include "FreeRTOSIPConfigDefaults.h"
~~~

Mentioned file has already been included.


htibosch:
Since [FreeRTOS-Plus-TCP](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP) has become a separate module, the development of FreeRTOS+TCP with multiple interfaces and IPv6, will be moved to that repo.
Closing this PR now.

PR #230, opened 2020-08-30
Address a few MISRA 2.2 violations in FreeRTOS_IP.c

AniruddhaKanhere:
Description
-----------
These changes bring in the changes required for addressing some of the MISRA rule 2.2 violations in FreeRTOS_IP.c.
The rule dictates that there shall be no dead code. When `ipconfigPRINTF` is not defined, then the debugging related variables seem to be unused. Use of ( void ) explicitly makes use of these variables while keeping the logic of code unchanged.

### DIFF
--- a/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
+++ b/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
@@ -2385,8 +2385,8 @@ BaseType_t location = 0;
 	{
 		FreeRTOS_printf( ( "CRC error: %04x location %ld\n", usChecksum, location ) );
 		
-		/* If FreeRTOS_printf is not defined, not using xLocation will be a violation of MISRA
-		 * rule 2.2 as the value assigned to location will not be used. The below statement uses
+		/* If FreeRTOS_printf is not defined, not using 'location' will be a violation of MISRA
+		 * rule 2.2 as the value assigned to 'location' will not be used. The below statement uses
 		 * the variable without modifying the logic of the source. */
 		( void ) location;
 	}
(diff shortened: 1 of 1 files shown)


FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
```diff
@@ -2105,6 +2105,7 @@ uint8_t ucProtocol;
 		if( xResult != pdPASS )
 		{
 			FreeRTOS_printf( ( "xCheckSizeFields: location %ld\n", xLocation ) );
+			( void ) xLocation;
```
gkwicker, line 2108: I would add a comment here as to why this statement exists; same for line 2383.

AniruddhaKanhere: Done, thanks!

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
```diff
@@ -2379,6 +2384,11 @@ BaseType_t location = 0;
 		( usChecksum == ipINVALID_LENGTH ) )
 	{
 		FreeRTOS_printf( ( "CRC error: %04x location %ld\n", usChecksum, location ) );
+		
+		/* If FreeRTOS_printf is not defined, not using xLocation will be a violation of MISRA
```
yanjos-dev, line 2388: Can you change "xLocation" to be "location" to match the variable name?

AniruddhaKanhere: I was supposed to fix that in the [previous commit](https://github.com/FreeRTOS/FreeRTOS/pull/230/commits/a3e02c19656da2732445356bc5bd45a0601f5b7b). Thanks for spotting that mistake.
It is fixed now.

PR #232, opened 2020-09-01
cbmc: Add patch to remove overflow assert

ravibhagavandas:
Add patch to comment overflow assert in queue proofs

Description
-----------
This is a  patch to comment the [assert check](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/805b15a0224672bc89b2effffb7ab4c1debc7325/queue.c#L398) for multiplication overflow in `xQueueGenericCreate`. CBMC performs large number of division computations as part of the check, which slows down the proof.  The proof does implicit check for signed and unsigned integer overflows.

Test Steps
-----------

Related Issue
-----------


n9wxu:
done

PR #240, opened 2020-09-02
Address various MISRA warnings v3

AniruddhaKanhere:
Description
-----------
This PR reduces the total violations by eliminating a few violations of different kinds.
MISRA rule 9.1, 10.8 violations addressed.


gkwicker:
The description mentions 21.15 and 8.6, however it doesn't look like this PR addresses either of these rules.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_Sockets.c
```diff
@@ -770,7 +770,7 @@ EventBits_t xEventBits = ( EventBits_t ) 0;
 			calculated at the total packet size minus the headers.
 			The validity of `xDataLength` prvProcessIPPacket has been confirmed
 			in 'prvProcessIPPacket()'. */
-			lReturn = ( int32_t ) ( pxNetworkBuffer->xDataLength - sizeof( UDPPacket_t ) );
+			lReturn = ( ( ( int32_t ) pxNetworkBuffer->xDataLength ) - ( ( int32_t ) sizeof( UDPPacket_t ) ) );
```
gkwicker, line 773: Which MISRA rule is this change addressing?  It is preferable to do unsigned arithmetic here and then cast to the signed value afterwards.  Same comments on the other 2 similar lines in this PR.

gkwicker (changes requested):
It's unclear which MISRA rule the changes to use signed arithmetic apply to; it's preferable to use unsigned arithmetic as was previously done and then cast to signed later.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_Sockets.c
AniruddhaKanhere: This addresses rule 10.8. Which essentially says `The value of composite expression shall not be cast to a different essential type or a wider essential type`.

gkwicker: The 10.8 violation is that the cast is being applied to the result of a composite expression.  We don't want to change this expression to use signed arithmetic since that changes the functionality.  Instead, do the following (verified in Coverity to address the 10.8 violation).  You can use this same approach to address the other violations in this PR rather than switching to use signed arithmetic.

```c
size_t xPayloadLength = 0;
.
.
.
    xPayloadLength = ( pxNetworkBuffer->xDataLength - sizeof( UDPPacket_t ) );
    lReturn = ( int32_t ) xPayloadLength;
```

gkwicker (changes requested):
See comment

gkwicker (approved):

tgsong (approved):

PR #242, opened 2020-09-04
TCP: Update history file and version number

AniruddhaKanhere:
Description
-----------
This PR updates the version number and updates the `History.txt` file for the +TCP source files.


gitwallit (dismissed):

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/History.txt
```diff
@@ -1,3 +1,47 @@
+Changes between V2.2.1 and V2.2.2
+	+ Add CBMC proofs for 4 functions.
```
RichardBarry, line 2: Folks may not know what CBMC is - so better to say something like "added another 4 memory safety proofs using the CBMC model checker".

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/History.txt
```diff
@@ -1,3 +1,47 @@
...
+	+ Address ~1200 MISRA violations.
```
RichardBarry, line 3: This looks really alarming, but I expect most of the 1200 are the same small change repeated in lots of places.  Best to be a bit more generic and say something like "stricter MISRA compliance", or "updated to conform to additional MISRA rules [then list them].

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/History.txt
```diff
@@ -1,3 +1,47 @@
...
+	+ Support caching multiple IP addresses per DNS entry in TCP library.
+	+ Security improvements in the TCP library.
```
RichardBarry, line 5: Is this in addition to the CBMC work?  Again I would make it either more generic as "continuous security improvements" or more specific such as "Improved security by giving users the option to reject UDP packets with a checksum of zero even though the spec allows it".

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/History.txt
```diff
@@ -1,3 +1,47 @@
...
+	+ Add additional length check when checksum offloading is used.
```
RichardBarry, line 6: "Improve robustness by checking packet lengths in software even if the check has already been performed in the hardware".

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/History.txt
```diff
@@ -1,3 +1,47 @@
...
+	+ Add ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS compile time constant to enable
+	  users to drop packets with IP options (IP Options are not supported).
+	+ Introduce ipconfigUDP_PASS_ZERO_CHECKSUM_PACKETS macro which enables the
```
RichardBarry, line 9: Say why, as per my previous comment "improve security by allows the user to drop UDP packets that have a checksum of zero even though the UDP specification says zero checksus should be accepted.".

RichardBarry (changes requested):

PR #246, opened 2020-09-08
FreeRTOS+TCP compatibility with 64-bit platforms

htibosch:
Description
-----------
In two functions, it was assumed that the size of a point is 32 bits, or less.
~~~c
NetworkBufferDescriptor_t *pxUDPPayloadBuffer_to_NetworkBuffer( const void * pvBuffer );
NetworkBufferDescriptor_t *pxPacketBuffer_to_NetworkBuffer( const void *pvBuffer );
~~~
With this patch, the functions can also be used on a 64-bit platform.

I tested the changes on a Xilinx Ultrascale+, which is 64-bits.


htibosch:
Two more changes related to do with 64-bit address operations in FreeRTOS_DNS.c :

~~~c
*pucStart = ( uint8_t ) ( ( uint32_t ) pucByte - ( uint32_t ) pucStart );
*pucStart = ( uint8_t ) ( ( uintptr_t ) pucByte - ( uintptr_t ) pucStart );

return ( ( uint32_t ) pucByte - ( uint32_t ) pucUDPPayloadBuffer + 1U ) + sizeof( DNSTail_t );
return ( ( uintptr_t ) pucByte - ( uintptr_t ) pucUDPPayloadBuffer + 1U ) + sizeof( DNSTail_t );
~~~

htibosch:
As the function `prvCreateDNSMessage()` is not too complicated, I will take away all pointer arithmetic and use an index.
Please ignore the last commit.
Hein

htibosch:
Just committed: "FreeRT_S_DNS.c remove pointer arithmetic in function `prvCreateDNSMessage()`".

The new version won't use pointers any more, it introduces index variables of type `size_t`. This approach is also more in line with the MISRA.
I tested the new code by comparing the output of the 2 versions of the function, both the contents and the length. The logic has not changed at all.

AniruddhaKanhere (approved):

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_DNS.c
```diff
@@ -828,11 +828,11 @@ TickType_t uxWriteTimeOut_ticks = ipconfigDNS_SEND_BLOCK_TIME_TICKS;
...
 
 _static size_t prvCreateDNSMessage( uint8_t *pucUDPPayloadBuffer,
-								   const char *pcHostName,
-								   TickType_t uxIdentifier )
+									const char *pcHostName,
+									TickType_t uxIdentifier )
 {
 DNSMessage_t *pxDNSMessageHeader;
-uint8_t *pucStart, *pucByte;
+size_t uxStart, uxIndex;
```
AniruddhaKanhere, line 835: While we are doing these changes, shall we also add a `configASSERT( pucUDPPayloadBuffer != NULL )`? Since that is directly used with the `memcpy` on line 849.
I would also like if we have a `configASSERT( pcHostName != NULL )`.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
```diff
@@ -895,47 +895,49 @@ NetworkBufferDescriptor_t * pxNewBuffer;
 }
 /*-----------------------------------------------------------*/
 
-#if( ipconfigZERO_COPY_TX_DRIVER != 0 ) || ( ipconfigZERO_COPY_RX_DRIVER != 0 )
+NetworkBufferDescriptor_t *pxPacketBuffer_to_NetworkBuffer( const void *pvBuffer )
+{
+uintptr_t uxBuffer;
+NetworkBufferDescriptor_t *pxResult;
```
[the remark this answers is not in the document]
htibosch: These functions:
~~~c
NetworkBufferDescriptor_t *pxPacketBuffer_to_NetworkBuffer( const void *pvBuffer );
NetworkBufferDescriptor_t *pxUDPPayloadBuffer_to_NetworkBuffer( const void * pvBuffer );
~~~
are very similar Each one subtracts a number of bytes from a pointer, and reads the address of a `NetworkBufferDescriptor_t`.
Because of this similarity, I propose to have the second call the first.

Another solution would be to add a common function:
~~~c
#if( ipconfigZERO_COPY_TX_DRIVER != 0 ) || ( ipconfigZERO_COPY_RX_DRIVER != 0 )
    NetworkBufferDescriptor_t *pxPacketBuffer_to_NetworkBuffer( const void *pvBuffer )
    {
        return prvBuffer_to_NetworkBuffer( pvBuffer, 0U );
    }
#endif /* ( ipconfigZERO_COPY_TX_DRIVER != 0 ) || ( ipconfigZERO_COPY_RX_DRIVER != 0 ) */

NetworkBufferDescriptor_t *pxUDPPayloadBuffer_to_NetworkBuffer( const void * pvBuffer )
{
    return prvBuffer_to_NetworkBuffer( pvBuffer, sizeof( UDPPacket_t ) );
}
~~~
Does that look better?

htibosch (commented):
Thanks for your review.
Hein

htibosch:
Commit "Merged two files with master, made changes suggested by Aniruddha"

I did the following:

- Added a new static function `prvPacketBuffer_to_NetworkBuffer` which is called from 2 locations
- Merged changes of `usGenerateChecksum()` from PR #286 
- Merged with the latest changes in master: `memcpy()` parameters

Since the last merge there are a few strange conflicts with master. I don't have the rights to solve them.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
```diff
@@ -945,26 +947,14 @@ NetworkBufferDescriptor_t *pxResult;
...
 	{
 		/* Obtain the network buffer from the zero copy pointer. */
-		pucBuffer = ipPOINTER_CAST( const uint8_t *, pvBuffer );
+		uxBuffer = ipPOINTER_CAST( uintptr_t, pvBuffer );
 
 		/* The input here is a pointer to a payload buffer.  Subtract
 		the total size of a UDP/IP header plus the size of the header in
 		the network buffer, usually 8 + 2 bytes. */
-		pucBuffer -= ( sizeof( UDPPacket_t ) + ( ( size_t ) ipBUFFER_PADDING ) );
+		uxBuffer -= ( uintptr_t ) sizeof( UDPPacket_t );
```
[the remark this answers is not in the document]
AniruddhaKanhere: Works for me. Thanks @htibosch.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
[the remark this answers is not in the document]
AniruddhaKanhere: Yes indeed. Both solutions would work.
I personally like the second soulution though.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_DNS.c
AniruddhaKanhere: I was just thinking from the perspective of CBMC checks. I would like to have proofs as unconstrained as possible. But if this configASSERT is a bit of an overstretch, then I am fine either way :)

AniruddhaKanhere:
@htibosch, thanks for the quick change to the PR. I have resolved and merged the conflicts. Do they seem okay to you?

AniruddhaKanhere (approved):

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_DNS.c
```diff
@@ -1333,7 +1333,7 @@ const size_t uxAddressLength = ipSIZE_OF_IPv4_ADDRESS;
 							 */
 							pvCopySource = &pucByte[ sizeof( DNSAnswerRecord_t ) ];
 							pvCopyDest = &ulIPAddress;
-							( void ) memcpy( pvCopyDest, pvCopySource, uxAddressLength );
+							( void ) memcpy( pvCopyDest, pvCopySource, sizeof( uint32_t ) );
```
htibosch: Oops, I did intend to use `uxAddressLength` in stead of `sizeof()`.
This has to do with the IPv6 release, in which `uxAddressLength` is variable ( 4 or 16 ).

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
```diff
@@ -926,7 +926,7 @@ NetworkBufferDescriptor_t *pxResult;
 			/* The following statement may trigger a:
 			warning: cast increases required alignment of target type [-Wcast-align].
 			It has been confirmed though that the alignment is suitable. */
-			pxResult = * ( ipPOINTER_CAST( NetworkBufferDescriptor_t **, uxBuffer ) );
+			pxResult = * ( ( NetworkBufferDescriptor_t ** ) pucBuffer );
```
htibosch: This was correct, `uxBuffer` is converted to a pointer-pointer.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_DNS.c
htibosch: I will add the following:
~~~c
	/* Althought both pointers have been checked already, two extra
	asserts are added to help the CBMC proofs.. */
	configASSERT( pucUDPPayloadBuffer != NULL );
	configASSERT( pcHostName != NULL );
~~~

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_DNS.c
AniruddhaKanhere: I see. But the `ulIPAddress` is defined as a 32 bit variable. And if we copy more than 4 bytes, do we not risk overwriting some other memory?
Is this intended? Or shall we be using some different datatype for `ulIPAddress`?

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_IP.c
AniruddhaKanhere: I think we still have `uxBuffer` in place. I see this in the file diff [here](https://github.com/FreeRTOS/FreeRTOS/pull/246/files#diff-a9fb45b06179576c1ce6f497e7012d46R929). Can you please check and verify?

htibosch: In your reply there is a [link](https://github.com/FreeRTOS/FreeRTOS/pull/246/files#diff-a9fb45b06179576c1ce6f497e7012d46R929) that shows correctly:
~~~c
    pxResult = * ( ( NetworkBufferDescriptor_t ** ) uxBuffer );
~~~
I have no idea where the conflicts with master have come from.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_DNS.c
htibosch: Why would we copy more than 4 bytes?
There is:
~~~c
    #define ipSIZE_OF_IPv4_ADDRESS     4U  /* See FreeRTOS_IP.h */
    const size_t uxAddressLength = ipSIZE_OF_IPv4_ADDRESS;
~~~
and so the `memcpy()` will copy 4 bytes only.

> Or shall we be using some different datatype for ulIPAddress
No, that is an ordinary IPv4 address, 4 bytes long.

AniruddhaKanhere (approved):

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_DNS.c
AniruddhaKanhere: I see. I was talking about this line from your comment "_This has to do with the IPv6 release, in which uxAddressLength is variable ( 4 or 16 )._"
What is 4 or 16? I assume that is the value in uxAddressLength. If it is 4, then it is fine, if it is 16, then we would be in trouble...

htibosch: As this is not interesting for the PR, I will reply by email.
Hein

htibosch:
This PR has moved to FreeRTOS-Plus-TCP [PR #20](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/20). Now closing this PR.

PR #271, opened 2020-09-15
Add reconnection and resubscribe logic to the MQTT plaintext demo

yourslab:
**Reconnection logic for the MQTT plaintext demo**

Description
-----------

This implements retries for FreeRTOS IoT Beta2 libraries, initializing the backoff to some value returned by the PRNG: `uxRand()`. If an attempt (connect or subscribe) fails, `vTaskDelay` is called to make the task sleep for some backoff period that doubles on each retry (but is bounded).

Test Steps
-----------

Just run the demo as you normally would.

Related Issue
-----------

### DIFF
--- a/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
+++ b/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
@@ -561,7 +561,7 @@ static void prvMQTTSubscribeWithBackoffRetries( MQTTContext_t * pxMQTTContext )
         LogInfo( ( "Attempt to subscribe to the MQTT topic %s.\r\n", mqttexampleTOPIC ) );
         xResult = MQTT_Subscribe( pxMQTTContext,
                                   &xGlobalSubscribeInfo,
-                                  sizeof( MQTTSubscribeInfo_t ),
+                                  sizeof( xGlobalSubscribeInfo ) / sizeof( MQTTSubscribeInfo_t ),
                                   usSubscribePacketIdentifier );
         configASSERT( xResult == MQTTSuccess );
 
@@ -632,8 +632,8 @@ static void prvMQTTUnsubscribeFromTopic( MQTTContext_t * pxMQTTContext )
      * was initialized before sending the SUBSCRIBE packet, there is no need
      * to initialize it again. */
     xResult = MQTT_Unsubscribe( pxMQTTContext,
-                                xGlobalSubscribeInfo,
-                                sizeof( MQTTSubscribeInfo_t ),
+                                &xGlobalSubscribeInfo,
+                                sizeof( xGlobalSubscribeInfo ) / sizeof( MQTTSubscribeInfo_t ),
                                 usUnsubscribePacketIdentifier );
 
     configASSERT( xResult == MQTTSuccess );
--- a/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
+++ b/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
@@ -266,7 +266,7 @@ static void prvMQTTProcessIncomingPublish( MQTTPublishInfo_t * pxPublishInfo );
 static void prvEventCallback( MQTTContext_t * pxMQTTContext,
                               MQTTPacketInfo_t * pxPacketInfo,
                               MQTTDeserializedInfo_t * pxDeserializedInfo );
-
+ 
 /*-----------------------------------------------------------*/
 
 /* @brief Static buffer used to hold MQTT messages being sent and received. */
@@ -653,7 +653,7 @@ static void prvMQTTPublishToTopic( MQTTContext_t * pxMQTTContext )
     /* Some fields not used by this demo so start with everything at 0. */
     ( void ) memset( ( void * ) &xMQTTPublishInfo, 0x00, sizeof( xMQTTPublishInfo ) );
 
-    /* This demo uses QOS0 */
+    /* This demo uses QOS0. */
     xMQTTPublishInfo.qos = MQTTQoS0;
     xMQTTPublishInfo.retain = false;
     xMQTTPublishInfo.pTopicName = mqttexampleTOPIC;
@@ -675,7 +675,9 @@ static void prvMQTTUnsubscribeFromTopic( MQTTContext_t * pxMQTTContext )
     /* Get next unique packet identifier */
     usUnsubscribePacketIdentifier = MQTT_GetPacketId( pxMQTTContext );
 
-    /* Send UNSUBSCRIBE packet. */
+    /* Send UNSUBSCRIBE packet. Note that because #pxGlobalSubscriptionList 
+     * was initialized before sending the SUBSCRIBE packet, there is no need
+     * to initialize it again. */
     xResult = MQTT_Unsubscribe( pxMQTTContext,
                                 pxGlobalSubscriptionList,
                                 sizeof( pxGlobalSubscriptionList ) / sizeof( MQTTSubscribeInfo_t ),
--- a/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
+++ b/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
@@ -351,10 +351,10 @@ static void prvMQTTDemoTask( void * pvParameters )
         /****************************** Connect. ******************************/
 
         /* Attempt to connect to the MQTT broker. If connection fails, retry after
-         * a timeout. Timeout value will be exponentially increased till the maximum
-         * attempts are reached or maximum timeout value is reached. The function
-         * returns a failure status if the TCP connection cannot be established to
-         * broker after configured number of attempts. */
+         * a timeout. Timeout value will be exponentially increased until the maximum
+         * number of attempts are reached or the maximum timeout value is reached.
+         * The function returns a failure status if the TCP connection cannot be established
+         * to the broker after the configured number of attempts. */
         xNetworkStatus = prvConnectToServerWithBackoffRetries( &xNetworkContext );
         configASSERT( xNetworkStatus == PLAINTEXT_TRANSPORT_SUCCESS );
 
--- a/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
+++ b/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
@@ -421,7 +421,6 @@ static PlaintextTransportStatus_t prvConnectToServerWithBackoffRetries( NetworkC
     RetryUtilsParams_t xReconnectParams;
 
     /* Initialize reconnect attempts and interval */
-    xReconnectParams.maxRetryAttempts = MAX_RETRY_ATTEMPTS;
     RetryUtils_ParamsReset( &xReconnectParams );
 
     /* Attempt to connect to MQTT broker. If connection fails, retry after
@@ -548,7 +547,6 @@ static void prvMQTTSubscribeWithBackoffRetries( MQTTContext_t * pxMQTTContext )
     xGlobalSubscribeInfo.topicFilterLength = ( uint16_t ) strlen( mqttexampleTOPIC );
 
     /* Initialize retry attempts and interval. */
-    xRetryParams.maxRetryAttempts = MAX_RETRY_ATTEMPTS;
     RetryUtils_ParamsReset( &xRetryParams );
 
     do
--- a/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
+++ b/FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
@@ -420,7 +420,7 @@ static PlaintextTransportStatus_t prvConnectToServerWithBackoffRetries( NetworkC
     RetryUtilsStatus_t xRetryUtilsStatus = RetryUtilsSuccess;
     RetryUtilsParams_t xReconnectParams;
 
-    /* Initialize reconnect attempts and interval */
+    /* Initialize reconnect attempts and interval. */
     RetryUtils_ParamsReset( &xReconnectParams );
 
     /* Attempt to connect to MQTT broker. If connection fails, retry after
@@ -481,7 +481,7 @@ static void prvCreateMQTTConnectionWithBroker( MQTTContext_t * pxMQTTContext,
     configASSERT( xResult == MQTTSuccess );
 
     /* Many fields not used in this demo so start with everything at 0. */
-    memset( ( void * ) &xConnectInfo, 0x00, sizeof( xConnectInfo ) );
+    ( void ) memset( ( void * ) &xConnectInfo, 0x00, sizeof( xConnectInfo ) );
 
     /* Start with a clean session i.e. direct the MQTT broker to discard any
      * previous session data. Also, establishing a connection with clean session
@@ -560,7 +560,7 @@ static void prvMQTTSubscribeWithBackoffRetries( MQTTContext_t * pxMQTTContext )
          * messages received from the broker will have QOS0. */
         LogInfo( ( "Attempt to subscribe to the MQTT topic %s.\r\n", mqttexampleTOPIC ) );
         xResult = MQTT_Subscribe( pxMQTTContext,
-                                  xGlobalSubscribeInfo,
+                                  &xGlobalSubscribeInfo,
                                   sizeof( MQTTSubscribeInfo_t ),
                                   usSubscribePacketIdentifier );
         configASSERT( xResult == MQTTSuccess );
@@ -625,7 +625,7 @@ static void prvMQTTUnsubscribeFromTopic( MQTTContext_t * pxMQTTContext )
 {
     MQTTStatus_t xResult;
 
-    /* Get next unique packet identifier */
+    /* Get next unique packet identifier. */
     usUnsubscribePacketIdentifier = MQTT_GetPacketId( pxMQTTContext );
 
     /* Send UNSUBSCRIBE packet. Note that because #xGlobalSubscribeInfo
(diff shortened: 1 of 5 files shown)


FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
```diff
@@ -305,15 +350,12 @@ static void prvMQTTDemoTask( void * pvParameters )
...
-        xNetworkStatus = Plaintext_FreeRTOS_Connect( &xNetworkContext,
-                                                     democonfigMQTT_BROKER_ENDPOINT,
-                                                     democonfigMQTT_BROKER_PORT,
-                                                     TRANSPORT_SEND_RECV_TIMEOUT_MS,
-                                                     TRANSPORT_SEND_RECV_TIMEOUT_MS );
+        /* Attempt to connect to the MQTT broker. If connection fails, retry after
+         * a timeout. Timeout value will be exponentially increased till the maximum
+         * attempts are reached or maximum timeout value is reached. The function
+         * returns a failure status if the TCP connection cannot be established to
+         * broker after configured number of attempts. */
```
muneebahmed10, line 357:
```suggestion
        /* Attempt to connect to the MQTT broker. If connection fails, retry after
         * a timeout. Timeout value will be exponentially increased until the maximum
         * number of attempts are reached or the maximum timeout value is reached.
         * The function returns a failure status if the TCP connection cannot be established
         * to the broker after the configured number of attempts. */
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
```diff
@@ -451,35 +552,93 @@ static void prvCreateMQTTConnectionWithBroker( MQTTContext_t * pxMQTTContext,
...
+    size_t pSize = 0;
+
+    xResult = MQTT_GetSubAckStatusCodes( pxPacketInfo, &pPayload, &pSize );
+
+    /* MQTT_GetSubAckStatusCodes always returns success if called with packet info
+     * from the event callback and non-NULL parameters. */
+    configASSERT( xResult == MQTTSuccess );
+
+    /* Demo only subscribes to one topic, so only one status code is returned. */
+    *pxGlobalSubAckStatus = pPayload[ 0 ];
```
muneebahmed10, line 596: I think we should be consistent in accessing pointers. It looks weird to me that `pxGlobalSubAckStatus` (declared as an array) is accessed with `*`, but `pPayload` (declared as a pointer) is accessed with `[]`. I think accessing both with `[]` would be the most logical

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
```diff
@@ -451,35 +552,93 @@ static void prvCreateMQTTConnectionWithBroker( MQTTContext_t * pxMQTTContext,
...
+
+    /* Initialize retry attempts and interval. */
+    RetryUtils_ParamsReset( &xRetryParams );
+
+    do
+    {
+        /* Send SUBSCRIBE packet.
+         * Note: reusing the value specified in globalSubscribePacketIdentifier is acceptable here
+         * because this function is entered only after the receipt of a SUBACK, at which point
+         * its associated packet id is free to use. */
```
muneebahmed10, line 614: Nit:
```suggestion
         * its associated packet id is free to reuse. */
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
```diff
@@ -451,35 +552,93 @@ static void prvCreateMQTTConnectionWithBroker( MQTTContext_t * pxMQTTContext,
...
 
     configASSERT( xResult == MQTTSuccess );
 }
 /*-----------------------------------------------------------*/
 
+static void prvUpdateSubAckStatus( MQTTPacketInfo_t * pxPacketInfo )
+{
+    MQTTStatus_t xResult = MQTTSuccess;
+    uint8_t * pPayload = NULL;
+    size_t pSize = 0;
```
muneebahmed10, line 587: Shouldn't these follow the FreeRTOS naming convention like the other functions?
```suggestion
    uint8_t * pucPayload = NULL;
    size_t ulSize = 0;
```

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
```diff
@@ -512,25 +671,14 @@ static void prvMQTTPublishToTopic( MQTTContext_t * pxMQTTContext )
...
     /* Get next unique packet identifier */
     usUnsubscribePacketIdentifier = MQTT_GetPacketId( pxMQTTContext );
-    /* Make sure the packet id obtained is valid. */
-    configASSERT( usUnsubscribePacketIdentifier != 0 );
 
     /* Send UNSUBSCRIBE packet. */
     xResult = MQTT_Unsubscribe( pxMQTTContext,
-                                xMQTTSubscription,
-                                sizeof( xMQTTSubscription ) / sizeof( MQTTSubscribeInfo_t ),
+                                pxGlobalSubscriptionList,
```
muneebahmed10, line 680: We could consider adding a comment that we don't need to re-initialize this since it will be identical to the info used for the subscribe

muneebahmed10 (approved):

muneebahmed10 (dismissed):

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
```diff
@@ -260,6 +293,18 @@ static uint16_t usSubscribePacketIdentifier;
...
+ * @brief Status of latest Subscribe ACK;
+ * it is updated every time the callback function processes a Subscribe ACK.
+ */
+static MQTTSubAckStatus_t pxGlobalSubAckStatus[ 1 ] = { MQTTSubAckFailure };
+
+/**
+ * @brief Array to keep subscription topics.
+ * Used to re-subscribe to topics that failed initial subscription attempts.
+ */
+static MQTTSubscribeInfo_t pxGlobalSubscriptionList[ 1 ];
```
dan4thewin, line 306: An array of a single element?  Would the demo be better if we used several topics to illustrate the concept?

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
```diff
@@ -343,6 +385,18 @@ static void prvMQTTDemoTask( void * pvParameters )
         xMQTTStatus = MQTT_ProcessLoop( &xMQTTContext, mqttexamplePROCESS_LOOP_TIMEOUT_MS );
         configASSERT( xMQTTStatus == MQTTSuccess );
 
+        /* Check if recent subscription request has been rejected. globalSubAckStatus is updated
+         * in eventCallback to reflect the status of the SUBACK sent by the broker. */
+        if( pxGlobalSubAckStatus[ 0 ] == MQTTSubAckFailure )
```
dan4thewin, line 390: Again, I think this might be a more valuable demo if several topics were used.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
```diff
@@ -343,6 +385,18 @@ static void prvMQTTDemoTask( void * pvParameters )
...
+        {
+            /* If server rejected the subscription request, attempt to resubscribe to topic.
+             * Attempts are made according to the exponential backoff retry strategy
+             * implemented in retryUtils. */
+            LogInfo( ( "Server rejected initial subscription request. Attempting to re-subscribe to topic %s.",
+                       mqttexampleTOPIC ) );
+            prvHandleResubscribe( &xMQTTContext );
```
dan4thewin, line 397: So, this is not the same function used to subscribe the first time...  does that mean one has backoff and the other does not?  Would it be better to have a single function that works in both cases?

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
```diff
@@ -391,6 +448,50 @@ static void prvMQTTDemoTask( void * pvParameters )
...
 /*-----------------------------------------------------------*/
 
+static PlaintextTransportStatus_t prvConnectToServerWithBackoffRetries( NetworkContext_t * pNetworkContext )
+{
+    PlaintextTransportStatus_t xNetworkStatus;
+    RetryUtilsStatus_t xRetryUtilsStatus = RetryUtilsSuccess;
+    RetryUtilsParams_t xReconnectParams;
+
+    /* Initialize reconnect attempts and interval */
+    RetryUtils_ParamsReset( &xReconnectParams );
```
dan4thewin, line 458: Could this functionality also be provided as a macro to use at declaration time?

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
```diff
@@ -543,7 +692,15 @@ static void prvMQTTProcessResponse( MQTTPacketInfo_t * pxIncomingPacket,
...
     {
         case MQTT_PACKET_TYPE_SUBACK:
-            LogInfo( ( "Subscribed to the topic %s.\r\n", mqttexampleTOPIC ) );
+
+            /* A SUBACK from the broker, containing the server response to our subscription request, has been received.
+             * It contains the status code indicating server approval/rejection for the subscription to the single topic
+             * requested. The SUBACK will be parsed to obtain the status code, and this status code will be stored in global
+             * variable globalSubAckStatus. */
+            prvUpdateSubAckStatus( pxIncomingPacket );
+            LogInfo( ( "Subscribed to the topic %s with maximum QoS %u.\r\n",
```
dan4thewin, line 701: Should this log message say something different for the 0x80 case?

FreeRTOS-Plus/Source/FreeRTOS-IoT-Libraries-LTS-Beta2/c_sdk/platform/freertos/retry_utils/retry_utils_freertos.c
```diff
@@ -0,0 +1,101 @@
...
+/*-----------------------------------------------------------*/
+
+RetryUtilsStatus_t RetryUtils_BackoffAndSleep( RetryUtilsParams_t * pRetryParams )
+{
+    RetryUtilsStatus_t status = RetryUtilsRetriesExhausted;
+    int32_t backOffDelayMs = 0;
+
+    /* If MAX_RETRY_ATTEMPTS is set to 0, try forever. */
+    if( ( pRetryParams->attemptsDone < MAX_RETRY_ATTEMPTS ) ||
+        ( 0 == MAX_RETRY_ATTEMPTS ) )
```
dan4thewin, line 49: This might be more useful if the max attempts lived in the struct.  Then you could default it to a value from a config, or override it directly from the calling function.

FreeRTOS-Plus/Source/FreeRTOS-IoT-Libraries-LTS-Beta2/c_sdk/platform/freertos/retry_utils/retry_utils_freertos.c
```diff
@@ -0,0 +1,101 @@
...
+    {
+        /* Choose a random value for back-off time between 0 and the max jitter value. */
+        backOffDelayMs = uxRand() % pRetryParams->nextJitterMax;
+
+        /*  Wait for backoff time to expire for the next retry. */
+        vTaskDelay( pdMS_TO_TICKS( backOffDelayMs * _MILLISECONDS_PER_SECOND ) );
```
dan4thewin, line 55: Seem to be conflating the backoff time and the jitter.  To my mind, this function should sleep for backoff + [0, max jitter], not [0, backoff + jitter].

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
yourslab: I think it would be best to just have this NOT be an array. We have another subscription manager demo to illustrate subscribing to multiple topics.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
yourslab: Good suggestion, I have made this to a single function. Thanks

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
yourslab: Good catch! I only print this now when `xGlobalSubAckStatus != MQTTSubAckFailure`. The warning message for the 0x80 is printed in the resubscription private function.

FreeRTOS-Plus/Source/FreeRTOS-IoT-Libraries-LTS-Beta2/c_sdk/platform/freertos/retry_utils/retry_utils_freertos.c
yourslab: While I do agree, I did not write this retry logic and am simply using the same implementation in CSDK. I am afraid that changing it might break something. Please feel free to chime in, @abhidixi11.

FreeRTOS-Plus/Demo/FreeRTOS-IoT-Libraries-LTS-Beta2/mqtt/mqtt_plain_text/DemoTasks/PlaintextMQTTExample.c
yourslab: Yes, but this is also how it is implemented in CSDK.

FreeRTOS-Plus/Source/FreeRTOS-IoT-Libraries-LTS-Beta2/c_sdk/platform/freertos/retry_utils/retry_utils_freertos.c
muneebahmed10: No, I think it's correct here. Sleeping for backoff + [0, max jitter] is closer to the "Equal jitter" approach described [here](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/), while the current implementation is closer to "Full jitter". In the blog, the equal jitter approach lost to full jitter in an experiment, in both completion time and work done.

FreeRTOS-Plus/Source/FreeRTOS-IoT-Libraries-LTS-Beta2/c_sdk/platform/include/retry_utils.h
```diff
@@ -202,6 +202,12 @@ typedef enum RetryUtilsStatus
  */
 typedef struct RetryUtilsParams
 {
+    /**
+     * @brief Max number of retry attempts. Set this value to 0 if the client must
+     * retry forever.
+     */
+    uint32_t maxRetryAttempts;
```
muneebahmed10, line 209: I don't think we should be making changes to this header file. Wouldn't it make more sense to keep in sync with the C SDK?

yourslab: The trouble is that the retry mechanism was built with the assumption that it would be only used for reconnection, but I imagine there will be cases where it will be useful to change the number of max attempts. Anyway, I will keep it in sync for now, but I think it is really something to consider. @dan4thewin

FreeRTOS-Plus/Source/FreeRTOS-IoT-Libraries-LTS-Beta2/c_sdk/platform/freertos/retry_utils/retry_utils_freertos.c
yourslab: Interesting article but how well do these simulations apply to things outside OCC? In the case of connecting to a server vs.  writing a database transaction, multiple clients would be able to succeed in the same round. It would be nice if the application could define the backoff algorithm (maybe for the well-architected SDK?). I imagine that some types of jitter work better for certain use cases than others.

leegeth (dismissed):

leegeth (approved):

muneebahmed10 (approved):

FreeRTOS-Plus/Source/FreeRTOS-IoT-Libraries-LTS-Beta2/c_sdk/platform/include/retry_utils.h
muneebahmed10: Wait, so this was added back?

FreeRTOS-Plus/Source/FreeRTOS-IoT-Libraries-LTS-Beta2/c_sdk/platform/freertos/retry_utils/retry_utils_freertos.c
dan4thewin: I'm also a bit skeptical about applying the conclusions of the article to our use cases.  Still, I think these are reasonable arguments for keeping this logic as is.

dan4thewin (approved):

FreeRTOS-Plus/Source/FreeRTOS-IoT-Libraries-LTS-Beta2/c_sdk/platform/freertos/retry_utils/retry_utils_freertos.c
muneebahmed10: Idk, but it is the article that we ourselves link to in this same file, so I thought it'd have to be at least somewhat relevant.

FreeRTOS-Plus/Source/FreeRTOS-IoT-Libraries-LTS-Beta2/c_sdk/platform/include/retry_utils.h
yourslab: @muneebahmed10 Sorry only saw this comment now. Yes, we added it back as it is not a part of any public-facing API that our users consume.

PR #274, opened 2020-09-16
Address MISRA Rule violations

gkwicker:
Description
-----------

1. FreeRTOS+TCP had many Rule 2.2 violations caused by the disabling of `inline` in Coverity builds.  This change eliminates this difference between Coverity and normal builds.
1. Local debugging trace variables were declared which were not used in the Coverity build because `ipconfigHAS_PRINTF` is not set in those builds.  This change disables these local debugging trace variables for this case.
1. Added an explanation for the inline suppression of Rule 11.4 in `prvSocketValid()`.

Test Steps
-----------

Related Issue
-----------


FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/include/FreeRTOS_Sockets.h
```diff
@@ -207,8 +207,19 @@ typedef struct xSOCKET const * ConstSocket_t;
...
+     * both values, the code cannot be compliant with rule 11.4,
+     * hence the Coverity suppression statement below.
+     */
     /* coverity[misra_c_2012_rule_11_4_violation] */
-    return ( ( xSocket != FREERTOS_INVALID_SOCKET ) && ( xSocket != NULL ) );
+    if( ( xSocket != FREERTOS_INVALID_SOCKET ) && ( xSocket != NULL ) )
+    {
+	    lReturnValue = pdTRUE;
+    }
+    return lReturnValue;
```
gkwicker, line 222: Explicitly declaring `lReturnValue` with the function return type eliminates a MISRA Rule violation.

AniruddhaKanhere (approved):

dan4thewin (approved):

PR #279, opened 2020-09-17
Update version number to V10.4.1

dachalco:
Update file header kernel version for `10.4.1`


cobusve (changes requested):
We need to update this one to exclude PR 272, 274 and 275

PR #280, opened 2020-09-17
FreeRTOS+TCP: MISRA rules 10.4, 10.8, & 21.15

gkwicker:
Description
-----------

These changes address FreeRTOS+TCP MISRA violations of rules 10.4, 10.8, and 21.15.

Test Steps
-----------

Related Issue
-----------


AniruddhaKanhere (dismissed):

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_DNS.c
```diff
@@ -1656,7 +1656,7 @@ BaseType_t xReturn;
 	/* This must be the first time this function has been called.  Create
 	the socket. */
 	xSocket = FreeRTOS_socket( FREERTOS_AF_INET, FREERTOS_SOCK_DGRAM, FREERTOS_IPPROTO_UDP );
-	if( prvSocketValid( xSocket ) != pdTRUE )
+	if( prvSocketValid( xSocket ) != pdTRUE_UNSIGNED )
```
AniruddhaKanhere, line 1659: I cannot find `pdTRUE_UNSIGNED` in projdefs.h. Where is this defined?

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_TCP_IP.c
```diff
@@ -967,8 +967,8 @@ static void prvTCPCreateWindow( FreeRTOS_Socket_t *pxSocket )
 	}
 	vTCPWindowCreate(
 		&pxSocket->u.xTCP.xTCPWindow,
-		( ( size_t ) ipconfigTCP_MSS ) * pxSocket->u.xTCP.uxRxWinSize,
-		( ( size_t ) ipconfigTCP_MSS ) * pxSocket->u.xTCP.uxTxWinSize,
+		ipconfigTCP_MSS * pxSocket->u.xTCP.uxRxWinSize,
+		ipconfigTCP_MSS * pxSocket->u.xTCP.uxTxWinSize,
```
AniruddhaKanhere, line 971: Would this not affect the functionality?

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_DNS.c
gkwicker: It's defined [here](https://github.com/FreeRTOS/FreeRTOS/blob/master/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/include/FreeRTOS_IP.h#L234).

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_TCP_IP.c
gkwicker: Typecasts _can_ affect the value of a variable, for example if casting to a smaller sized type, between `float` and `int` types, or between signed and unsigned types when dealing with values in which the high bit is set.  But because [we know the range of possible values](https://en.wikipedia.org/wiki/Maximum_transmission_unit) for [`ipconfigTCP_MSS`](https://github.com/FreeRTOS/FreeRTOS/blob/master/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/include/FreeRTOSIPConfigDefaults.h#L393) (an integer constant), we know that none of these conditions exist here.

FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_DNS.c
AniruddhaKanhere: Thanks!

AniruddhaKanhere (approved):

aggarg (approved):

PR #286, opened 2020-09-20
FreeRTOS+TCP Fix usGenerateChecksum on 8-bit platforms with odd-aligned buffers

andyjpb:
Fix usGenerateChecksum on 8-bit platforms with odd-aligned buffers

This is a resubmission of https://github.com/FreeRTOS/FreeRTOS/pull/122, requested by Hien Tibosch as this fix is required for some of their new proposals.

This has been tested on 8-bit AVR and 64-bit Xilinx UltraScale+.

This pull request includes the patches from #122 as well as a patch that covers the feedback regarding using a uintptr_t rather than an intptr_t.

-----------

Description
-----------
 + Fix casting of pointers to integers when pointers aren't 32-bits wide.
 + Fix generation of checksum on odd-aligned buffers with non-zero initialisation vector.

Test Steps
-----------
On avr8 the stack complains about invalid checksums for arriving UDP and TCP packets until this patch is applied.

Surprisingly, the stack can receive the DHCP packets necessary to obtain an IP address but all other UDP and TCP packets seem to fail the UDP or TCP checksum. The IP header checksum is not affected.

Reproduce:
TCP:
...by trying to telnet to a port on the target
`$ telnet 172.25.11.35 51`

UDP:
...by using netcat to send packets to the target
`echo "Hello" | nc -u 172.25.11.35 51`

I was running this on an ATMega1284 with an Microchip ENC28J60 combined MAC and PHY.

I reproduced the issue by compiling the usGenerateChecksum function on Linux-buster-x86_64 and passing it captured packets that I had pre-padded to be aligned on different boundaries. I can supply this code if desired.


AniruddhaKanhere (approved):

htibosch:
Thanks Andy for this PR. You discovered that `usGenerateChecksum()` comes with a wrong answer as soon as the packet buffer is byte-aligned, You also found the solution.

Normally, on a 32-bit platform, the alignment will be either 0 or 2. You were using an ATmega, which is essentially 8-bits.

I just linked your new version of `FreeRTOS_IP.c` in a testing project ( STM32F4x, little endian ), and had packets checked with all for4 alignments: 0, 1, 2, 3. The outputs for both `usGenerateChecksum()` and `usGenerateProtocolChecksum` were correct in all cases.
Earlier I tested it on a big-endian platform UC3.

Beside that we solved a minor MISRA issue: a void pointer will now be converted to `uintptr_t` in stead of a `uint32_t`, which may have the wrong size.

Hein

andyjpb:
Hi,

Thanks for the detailed summary, Hein!

I have been happy to contribute and thanks for supporting my pull request 
all the way through.


> Thanks Andy for this PR. You discovered that 
> usGenerateChecksum() comes with a wrong answer as soon as the 
> packet buffer is byte-aligned, You also found the solution.
>
> Normally, on a 32-bit platform, the alignment will be either 0 
> or 2. You were using an ATmega, which is essentially 8-bits.
>
> I just linked your new version of FreeRTOS_IP.c in a testing 
> project ( STM32F4x, little endian ), and had packets checked 
> with all for4 alignments: 0, 1, 2, 3. The outputs for both 
> usGenerateChecksum() and usGenerateProtocolChecksum were correct 
> in all cases.
> Earlier I tested it on a big-endian platform UC3.
>
> Beside that we solved a minor MISRA issue: a void pointer will 
> now be converted to uintptr_t in stead of a uint32_t, which may 
> have the wrong size.
>
> Hein
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.

-- 
Best wishes,
@ndy

-- 
andyjpb@ashurst.eu.org
http://www.ashurst.eu.org/
0x7EBA75FF

htibosch:
If anyone else could give a second approval, I can carry on with my [PR #246](https://github.com/FreeRTOS/FreeRTOS/pull/246), that depends on this one.

Thanks, Hein

RichardBarry (approved):

htibosch:
@AniruddhaKanhere : thanks for your efforts!
Hein

PR #313, opened 2020-10-01
Support simultaneous connections in mbedTLS transport wrapper

yourslab:
Description
-----------

When calling `TLS_FreeRTOS_Disconnect`, `mbedtls_entropy_free` and `mbedtls_ctr_drbg_free` are called. If `TLS_FreeRTOS_Send` or `TLS_FreeRTOS_Recv` is called afterwards from another connection, then it will fail because the entropy context has been freed. This adds the entropy context as part of the network context object so that each connection has one to prevent that.


yourslab:
Closing this as the changes are added to PR #337

PR #329, opened 2020-10-08
Remove extraneous call to MQTT_ProcessLoop

aggarw13:
Remove unncessary call to `MQTT_Process﻿Loop` after the `prvMQTTSubscribeWithBackoffRetries` call in MQTT mutual auth demo. 

Call to `MQTT_ProcessLoop` is not required as the `prvMQTTSubscribeWithBackoffRetries` private function _does_ call `MQTT_ProcessLoop` function (to receive SUBACK response from broker)


yourslab:
Thanks. Just merged #331, can you do this against `master` now?

yourslab (approved):

n9wxu (approved):

PR #11, opened 2020-02-12
Updates to Markdown files and readme.txt

cobusve:

PR #338, opened 2020-10-12
Move PKCS #11 Mutual Auth Demo from LTS Dev branch

lundinc2:
Move PKCS #11 Mutual Auth Demo from LTS Dev branch

Description
-----------
This change moves the PKCS #11 Mutual Auth demo. The demo has previously been reviewed, but the project needed to be updated to address changes to the folder structure.

Test Steps
-----------
Follow the demo instructions on freertos.org for the PKCS #11 mutual auth demo, and connect to AWS IoT Core.

Related Issue
-----------


yourslab (approved):
Nit: Maybe a better folder name for this demo would be `MQTT_PKCS11_Mutual_Auth_Windows_Simulator` because we might have to change the other demo names to `coreMQTT_MQTT_Mutual_Auth_Windows_Simulator` for consistency and that just doesn't look right. Although, this name may have been decided in the past already.

lundinc2:
> Nit: Maybe a better folder name for this demo would be `MQTT_PKCS11_Mutual_Auth_Windows_Simulator` because we might have to change the other demo names to `coreMQTT_MQTT_Mutual_Auth_Windows_Simulator` for consistency and that just doesn't look right. Although, this name may have been decided in the past already.

Thanks, this is a good thought. For now I prefer this. The other demos share a lot of code and I think it makes sense for them. In this case, I kept the PKCS mutual auth demo separate from the other ones, and is it has a lot more dependencies / more complex overall

gkwicker (approved):

PR #340, opened 2020-10-13
Delete all the guidance to use the public mosquitto broker.

sarenameas:
* We do not have service level guarantees on the public broker, so it not recommended to have customers use the third party host.
* We tell customers they can use the public broker, but then we tell them that it is unstable. It is better to not tell customers at all to use it.


yourslab (approved):

RichardBarry:
The instructions for the plain text demo connects to Mosquitto (https://www.freertos.org/mqtt/preconfiguredexamples.html), but we need words that describes how this is done to experiment with the MQTT protocol to make sure body sends private data and to make it is clear that production devices should never use an unencrypted connection (as per the red text on the link).

gkwicker (approved):

PR #362, opened 2020-10-23
Improve connection sharing demo

muneebahmed10:
Description
-----------

Makes changes to the connection sharing demo after feedback. Changes include:
* Add a daemon socket listener so that the process loop will only execute if data is available on the socket.
* Set process loop timeout to 0 so that other commands do not have to wait for a process loop that isn't doing anything.
* Creates an additional publisher task to show multiple tasks publishing simultaneously.
* Clarify what is meant by "synchronous" and "asynchronous" publishes.
* Fix possible race condition in task notification waits.
* Move task notification wait loops to its own function.
* The `prvMQTTConnect` function was doing more than just connecting, so it was split into more functions.
* Minor name changes for clarity.

Test Steps
-----------

Tested locally against mosquitto and the AWS IoT Broker.

### DIFF
--- a/FreeRTOS-Plus/Demo/coreMQTT_Windows_Simulator/MQTT_Multitask/DemoTasks/MultitaskMQTTExample.c
+++ b/FreeRTOS-Plus/Demo/coreMQTT_Windows_Simulator/MQTT_Multitask/DemoTasks/MultitaskMQTTExample.c
@@ -108,12 +108,29 @@
     #ifndef democonfigROOT_CA_PEM
         #error "Please define Root CA certificate of the MQTT broker(democonfigROOT_CA_PEM) in demo_config.h."
     #endif
-    #ifndef democonfigCLIENT_CERTIFICATE_PEM
-        #error "Please define client certificate(democonfigCLIENT_CERTIFICATE_PEM) in demo_config.h."
-    #endif
-    #ifndef democonfigCLIENT_PRIVATE_KEY_PEM
-        #error "Please define client private key(democonfigCLIENT_PRIVATE_KEY_PEM) in demo_config.h."
-    #endif
+
+/* If no username is defined, then a client certificate/key is required. */
+    #ifndef democonfigCLIENT_USERNAME
+        #ifndef democonfigCLIENT_CERTIFICATE_PEM
+            #error "Please define client certificate(democonfigCLIENT_CERTIFICATE_PEM) in demo_config.h."
+        #endif
+        #ifndef democonfigCLIENT_PRIVATE_KEY_PEM
+            #error "Please define client private key(democonfigCLIENT_PRIVATE_KEY_PEM) in demo_config.h."
+        #endif
+    #else
+
+/* If a username is defined, a client password also would need to be defined for
+ * client authentication. */
+        #ifndef democonfigCLIENT_PASSWORD
+            #error "Please define client password(democonfigCLIENT_PASSWORD) in demo_config.h for client authentication based on username/password."
+        #endif
+
+/* AWS IoT MQTT broker port needs to be 443 for client authentication based on
+ * username/password. */
+        #if defined( democonfigUSE_AWS_IOT_CORE_BROKER ) && democonfigMQTT_BROKER_PORT != 443
+            #error "Broker port(democonfigMQTT_BROKER_PORT) should be defined as 443 in demo_config.h for client authentication based on username/password in AWS IoT Core."
+        #endif
+    #endif /* ifndef democonfigCLIENT_USERNAME */
 
     #ifndef democonfigMQTT_BROKER_PORT
         #define democonfigMQTT_BROKER_PORT    ( 8883 )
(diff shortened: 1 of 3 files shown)


FreeRTOS-Plus/Demo/coreMQTT_Windows_Simulator/MQTT_Multitask/DemoTasks/MultitaskMQTTExample.c
```diff
@@ -751,21 +826,36 @@ static MQTTStatus_t prvMQTTConnect( MQTTContext_t * pxMQTTContext,
...
+    }
+
+    return xResult;
+}
+
+/*-----------------------------------------------------------*/
+
+static void prvResumeSession( bool xSessionPresent )
+{
+    MQTTStatus_t xResult;
```
yourslab, line 858: This variable is assigned the status of the publish but it never seems to be used beyond that. Could it simply be removed or maybe queuing it in the event of failure?

FreeRTOS-Plus/Demo/coreMQTT_Windows_Simulator/MQTT_Multitask/DemoTasks/MultitaskMQTTExample.c
```diff
@@ -908,14 +997,14 @@ static BaseType_t prvConnectNetwork( NetworkContext_t * pxNetworkContext )
                                                    &xNetworkCredentials,
                                                    mqttexampleTRANSPORT_SEND_RECV_TIMEOUT_MS,
                                                    mqttexampleTRANSPORT_SEND_RECV_TIMEOUT_MS );
-            xConnected = ( xNetworkStatus == TLS_TRANSPORT_SUCCESS );
+            xConnected = ( xNetworkStatus == TLS_TRANSPORT_SUCCESS ) ? true : false;
```
yourslab, line 1000: Would it be better to use a `BaseType_t`?

FreeRTOS-Plus/Demo/coreMQTT_Windows_Simulator/MQTT_Multitask/DemoTasks/MultitaskMQTTExample.c
```diff
@@ -1779,12 +1944,14 @@ void prvSubscribeTask( void * pvParameters )
          * the last publish. */
         if( ++ulWaitCounter > mqttexampleMAX_WAIT_ITERATIONS )
         {
-            LogError( ( "Publish receive loop exceeded maximum wait time." ) );
+            LogError( ( "Publish receive loop exceeded maximum wait time.\n" ) );
```
yourslab, line 1947: Why does it make sense to log a `\n` here but not everywhere else?

FreeRTOS-Plus/Demo/coreMQTT_Windows_Simulator/MQTT_Multitask/DemoTasks/MultitaskMQTTExample.c
```diff
@@ -92,7 +101,7 @@
 /* Compile time error for some undefined configs, and provide default values
  * for others. */
 #ifndef democonfigMQTT_BROKER_ENDPOINT
-    #error "Please define democonfigMQTT_BROKER_ENDPOINT."
+    #error "Please define democonfigMQTT_BROKER_ENDPOINT in demo_config.h."
```
yourslab, line 104: If this is intended to work with AWS IoT, then supporting username/password along with metrics might be a good idea.

muneebahmed10: I didn't want this demo to do too much since we have the mutual auth demo for that, though I did plan on using this with AWS IoT. How much work would that be?

FreeRTOS-Plus/Demo/coreMQTT_Windows_Simulator/MQTT_Multitask/DemoTasks/MultitaskMQTTExample.c
muneebahmed10: So I had intended some log messages to be closer together if they're related (e.g. "Starting wait" and "Wait complete") and others use newlines for spacing. Since error logs aren't normally supposed to happen, I added newlines here. I'll add them to the other error logs too.

FreeRTOS-Plus/Demo/coreMQTT_Windows_Simulator/MQTT_Multitask/DemoTasks/MultitaskMQTTExample.c
muneebahmed10: Might as well. I tried to stick with either `bool` or `BaseType_t` consistently and settled on `bool` since that's what MQTT uses, but since the return type for this function was changed a while back I can just convert the `bool`s here

FreeRTOS-Plus/Demo/coreMQTT_Windows_Simulator/MQTT_Multitask/DemoTasks/MultitaskMQTTExample.c
muneebahmed10: I didn't check it since I didn't want a failed publish to stop the demo, but it probably should. I think I'll change the return type for the function so if there's an error, the loop will disconnect and try the connection again.

FreeRTOS-Plus/Demo/coreMQTT_Windows_Simulator/MQTT_Multitask/DemoTasks/MultitaskMQTTExample.c
```diff
@@ -1544,39 +1696,59 @@ static void prvCommandCallback( CommandContext_t * pxContext )
...
 /*-----------------------------------------------------------*/
 
-void prvPublishTask( void * pvParameters )
+static bool prvNotificationWaitLoop( uint32_t * pulNotification,
+                                     uint32_t ulExpectedBits,
+                                     bool xClearBits )
+{
+    uint32_t ulWaitCounter = 0U;
+    bool ret = true;
+
```
yourslab, line 1705: We may want to assert that `pulNotification != NULL`

FreeRTOS-Plus/Demo/coreMQTT_Windows_Simulator/MQTT_Multitask/DemoTasks/MultitaskMQTTExample.c
yourslab: Just that I remember we wanted to have mutual auth demos automatically picking up metrics. Just requires copying and pasting a few macros and adding them to the connect info.

yourslab (approved):

n9wxu (approved):

PR #384, opened 2020-11-06
Allow overriding default definitions of logging macros

aggarw13:
Update logging_stack.h to allow:
* User to provide definition to `SdkLog` ( instead of only defining the macro to Windows simulator specific `vLoggingPrintf` function). Also, move mapping of `SdkLog` to `vLoggingPrintf` to demo config files to be more expressive in demos about example of how user's can define `SdkLog` to their platform-specific logging function to enable logging.
* Allow users to custom define the metadata for logs through the `LOG_METADATA_ARGS` and `LOG_METADATA_FORMAT` macros.


FreeRTOS-Plus/Source/Utilities/logging/logging_stack.h
```diff
@@ -37,60 +37,91 @@
...
+/* Check if LIBRARY_LOG_NAME macro has been defined. */
+#if !defined( LIBRARY_LOG_NAME )
+    #error "Please define LIBRARY_LOG_NAME for the library."
+#endif
 
 /* Metadata information to prepend to every log message. */
-#define LOG_METADATA_FORMAT    "[%s:%d] "
-#define LOG_METADATA_ARGS      __FUNCTION__, __LINE__
+#ifndef LOG_METADATA_FORMAT
+    #define LOG_METADATA_FORMAT    "[%s] [%s:%d] "                  /**< @brief Format of metadata prefix in log messages as `[<Logging-Level>] [<Library-Name>] [<File-Name>:<Line-Number>]` */
```
yourslab, line 65: Still says File-Name in the comment rather than Function-Name

aggarw13, line 65:
```suggestion
    #define LOG_METADATA_FORMAT    "[%s] [%s:%d] "                  /**< @brief Format of metadata prefix in log messages as `[<Logging-Level>] [<Library-Name>] [<Function-Name>:<Line-Number>]` */
```

yourslab (dismissed):

yourslab (dismissed):

FreeRTOS-Plus/Source/Utilities/logging/logging_stack.h
```diff
@@ -37,60 +37,92 @@
...
+ */
 /* Check that LIBRARY_LOG_LEVEL is defined and has a valid value. */
 #if !defined( LIBRARY_LOG_LEVEL ) ||       \
     ( ( LIBRARY_LOG_LEVEL != LOG_NONE ) && \
     ( LIBRARY_LOG_LEVEL != LOG_ERROR ) &&  \
     ( LIBRARY_LOG_LEVEL != LOG_WARN ) &&   \
     ( LIBRARY_LOG_LEVEL != LOG_INFO ) &&   \
-    ( LIBRARY_LOG_LEVEL != LOG_DEBUG ) )
+    ( LIBRARY_LOG_LEVEL != LOG_DEBUG )     \
+    )
```
yourslab, line 99: Nit: this is formatted weird

yourslab (dismissed):

FreeRTOS-Plus/Source/Utilities/logging/logging_stack.h
```diff
@@ -37,25 +37,58 @@
 #include "logging_levels.h"
 
 /* Standard Include. */
+#include <stdio.h>
 #include <stdint.h>
 
+/* The macro definition for LIBRARY_LOG_NAME is for Doxygen
```
RichardBarry, line 43: I don't understand this comment as it implies the macro is only used by Doxygen, but the comment below says the macro is also used in log messages.

FreeRTOS-Plus/Source/Utilities/logging/logging_stack.h
```diff
@@ -37,25 +37,58 @@
...
+ */
+/* Check if LIBRARY_LOG_NAME macro has been defined. */
+#if !defined( LIBRARY_LOG_NAME )
+    #error "Please define LIBRARY_LOG_NAME for the library."
+#endif
+
 /* Metadata information to prepend to every log message. */
-#define LOG_METADATA_FORMAT    "[%s:%d] "
-#define LOG_METADATA_ARGS      __FUNCTION__, __LINE__
+#ifndef LOG_METADATA_FORMAT
```
RichardBarry, line 63: It looks like this is something users can define - if so it needs adding to the library documentation page.

FreeRTOS-Plus/Source/Utilities/logging/logging_stack.h
```diff
@@ -37,25 +37,58 @@
...
+#ifndef LOG_METADATA_ARGS
+    #define LOG_METADATA_ARGS    LIBRARY_LOG_NAME, __FUNCTION__, __LINE__ /**< @brief Arguments into the metadata logging prefix format. */
+#endif
 
 /* Prototype for the function used to print out.  In this case it prints to the
  * console before the network is connected then a UDP port after the network has
  * connected. */
-    extern void vLoggingPrintf( const char * pcFormatString,
-                                ... );
+extern void vLoggingPrintf( const char * pcFormatString,
```
RichardBarry, line 74: If this is a generic header file, intended for use with all FreeRTOS ports, then this line and the above comment shouldn't be here because they are specific to running FreeRTOS on Windows.

FreeRTOS-Plus/Source/Utilities/logging/logging_stack.h
```diff
@@ -37,25 +37,58 @@
...
+                            ... );
+
+/**
+ * @brief Common macro that maps all the logging interfaces,
+ * (#LogDebug, #LogInfo, #LogWarn, #LogError) to the platform-specific logging
+ * function.
+ *
+ * @note The default definition of the macro maps to the vLoggingPrintf function.
+ */
+#ifndef SdkLog
```
RichardBarry, line 84: This risks creating confusing linker errors if somebody uses this header file but does not include the Windows version of the logging macro.  I would suggest the default behavior should be to define this as an empty macro so it doesn't generate any output.

RichardBarry (changes requested):

FreeRTOS-Plus/Source/Utilities/logging/logging_stack.h
aggarw13: Good suggestion. Will make the change.

FreeRTOS-Plus/Source/Utilities/logging/logging_stack.h
aggarw13: Yes, will add to the library documentation page

FreeRTOS-Plus/Source/Utilities/logging/logging_stack.h
aggarw13: Makes sense, removed the prototype from this generic header.

FreeRTOS-Plus/Source/Utilities/logging/logging_stack.h
aggarw13: This comment was a copy-paste from the [C-SDK version of logging_stack.h](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/master/demos/logging-stack/logging_stack.h#L39-L60) which contains special text for doxygen generation. I have removed this comment about doxygen generation as it is not relevant in this repository.

FreeRTOS-Plus/Source/Utilities/logging/logging_stack.h
```diff
@@ -37,25 +37,49 @@
 #include "logging_levels.h"
```
yourslab, line 37: nit: I think the header include guards conventionally don't have the trailing underscore in FreeRTOS

aggarw13: Changed to `LOGGING_STACK`

yourslab (approved):

PR #391, opened 2020-11-08
Fix logging issue causing mangled new lines

yourslab:
Description
-----------

The Windows implementation of `vLoggingPrintf` used `\n` as the default new-line character. However, the demos used `\r\n`, so it would append the extra parameters such as `pcTaskName` to the end of each new-line message. This fixes it by setting the default new-line character to `\r\n`.

In addition, any `configTASKPOOL` macros are removed as they are not being used at all.

### DIFF
--- a/FreeRTOS-Plus/Demo/Common/Logging/windows/Logging_WinSim.c
+++ b/FreeRTOS-Plus/Demo/Common/Logging/windows/Logging_WinSim.c
@@ -71,9 +71,9 @@
 /* A block time of zero simply means don't block. */
 #define dlDONT_BLOCK                    0
 
-/* Choose LF as the default new-line character. */
+/* Choose CRLF as the default new-line character. */
 #ifndef configLOGGING_NEW_LINE
-    #define configLOGGING_NEW_LINE    "\n"
+    #define configLOGGING_NEW_LINE    "\r\n"
 #endif
 
 /*-----------------------------------------------------------*/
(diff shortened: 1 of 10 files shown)


FreeRTOS-Plus/Demo/Common/Logging/windows/Logging_WinSim.c
```diff
@@ -71,6 +71,11 @@
 /* A block time of zero simply means don't block. */
 #define dlDONT_BLOCK                    0
 
+/* Choose LF as the default new-line character. */
+#ifndef configLOGGING_NEW_LINE
+    #define configLOGGING_NEW_LINE    "\n"
```
RichardBarry, line 76: As the file is called Logging_WinSim - shouldn't the default be "\r\n"?

RichardBarry: Looks like if the above were the case the other changes would not be required.  Although taking out the task pool constants is desirable.

RichardBarry (changes requested):

FreeRTOS-Plus/Demo/Common/Logging/windows/Logging_WinSim.c
yourslab: Agreed. As this is used by other WinSim demos besides MQTT, I was a little worried about them using '\n' as the default. Upon further inspection, this was only true for `FreeRTOS_Plus_TCP_Minimal_Windows_Simulator`, so I updated that demo to use CRLF instead and have set the default to CRLF as well.

muneebahmed10 (approved):

RichardBarry (approved):

PR #403, opened 2020-11-11
Fix conflicting definition of SdkLog from config files

aggarw13:
#### Problem 
When there are different definitions of `SdkLog` in different config files (like `demo_config.h` and `core_mqtt_config.h` files) and the config files are included in the same translation unit, there is compiler warning of conflicting definitions. 

#### Fix 
Add `#ifndef` guards around the `SdkLog` definitions in config files to avoid multiple definitions in a translation unit. 

#### Note 
For most cases, the user may need to define the `SdkLog` macro ONLY once for the entire project build. An alternative way to define the macro for the whole project is by setting its default value in the `FreeRTOS-Plus/Source/Utilities/Logging/logging_stack.h` file

### DIFF
--- a/FreeRTOS-Plus/Demo/AWS/Device_Shadow_Windows_Simulator/Device_Shadow_Demo/core_mqtt_config.h
+++ b/FreeRTOS-Plus/Demo/AWS/Device_Shadow_Windows_Simulator/Device_Shadow_Demo/core_mqtt_config.h
@@ -56,7 +56,9 @@ extern void vLoggingPrintf( const char * pcFormatString,
 
 /* Map the SdkLog macro to the logging function to enable logging
  * on Windows simulator. */
-#define SdkLog( message )    vLoggingPrintf message
+#ifndef SdkLog
+    #define SdkLog( message )    vLoggingPrintf message
+#endif
 
 #include "logging_stack.h"
 /************ End of logging configuration ****************/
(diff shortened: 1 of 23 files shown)


abhidixi11 (approved):

RichardBarry:
It is important a user knows which definition of SdkLog() is going to be used in each library, and I don't think that is going to be the case if any particular definition of SdkLog() is only used if the compiler has not seen a prior definition.  How does the application writer know if the compiler has seen a prior definition?

> An alternative way to define the macro for the whole project is by setting its default value in the FreeRTOS-Plus/Source/Utilities/Logging/logging_stack.h file

logging_stack.h is within the source directory try rather than the demo directory tree.  That means its code can't change on a project by project basis.  There could only be a default value by #defining it to a function name, then asking the application writer to provide an implementation of that function as part of their application code (only the user knows what an appropriate implementation is).

aggarw13:
>It is important a user knows which definition of SdkLog() is going to be used in each library, and I don't think that is going to be the case if any particular definition of SdkLog() is only used if the compiler has not seen a prior definition. How does the application writer know if the compiler has seen a prior definition?

With the change in this PR, each library will still use the definition of `SdkLog` mentioned in the library-specific config file. That is, coreMQTT library will still use the definition for the macro provided in `core_mqtt_config.h` file, whereas, the demo files, like `MutualAuthMQTTExample.c`, will use the macro definition provided in `demo_config.h`. 
The only purpose of adding the `#ifndef` guards in the config files is that for cases when there are multiple config files included in the same translation unit (or .c file) like the `MutualAuthMQTTExample.c` (where it includes `demo_config.h` directly and `core_mqtt_config.h` indirectly through `core_mqtt.h` inclusion), there are no multiple definitions of the `LIBRARY_LOG_NAME`, `LIBRARY_LOG_LEVEL` and `SdkLog` macros.
However, for such cases, there is a header include order requirement that the config file relevent to the translation unit _is_ the first config file to be included in the file. Thus, `demo_config.h` header is the first config file to included in the demo source files so that the logging configurations of `demo_config.h` file is applied (instead of a library-specific config file like `core_mqtt_config.h`).

RichardBarry:
Ok - lets add the same guards for the other logging macros then update the logging documentation page to describe the order in which application writers include the configuration files in their own code.

RichardBarry (approved):

PR #408, opened 2020-11-17
File Header Check/Diff Reporter + Companion Git Action

dachalco:
Description
-----------
+ New script that checks file headers for input file(s). Shows familiar diff. Can accept list of files in `*.json` format too.
+ Git action that is triggered by pull requests. It can be added as a mandatory check so we can enforce file headers (version, copyright, licensing)

Test Steps
-----------
Run the script locally. See `check-header.py -h`.
Can also trigger git action by adding new PR.

Related Issue
-----------
n/a

### DIFF
--- a/.github/scripts/check-header.py
+++ b/.github/scripts/check-header.py
@@ -1,0 +1,122 @@
+#!/usr/bin/env python3
+
+import os, sys
+from argparse import ArgumentParser
+from difflib import unified_diff
+
+def dprint(msg):
+    print('[DEBUG]: %s' % str(msg))
+
+class HeaderChecker:
+    def __init__(self, header, prefix_tolerance=10):
+        self.prefix_tolerance = 10
+        self.header = header
+
+    def normalizeHeader():
+        assert False, 'Unimplemented'
+
+    def isValidFile(self, path):
+        assert os.path.exists(path), 'No such file: ' + path
+
+        # Don't need entire file. Read sufficienly large chunk of file that should contain the header
+        with open(path, encoding='utf-8', errors='ignore') as file:
+            chunk = file.read(len('\n'.join(self.header)))
+            lines = [('%s\n' % l) for l in chunk.strip().splitlines()]
+            if self.header == lines:
+                return True
+            else:
+                print('File Delta: %s' % path)
+                print(*unified_diff(lines[:len(self.header)], self.header))
+                return False
+
+
+def configArgParser():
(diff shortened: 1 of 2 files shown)


dachalco:
FYI, it's failing it's own check as it tries to run `FreeRTOS/FreeRTOS/.github/scripts/check-header.py` which, of course, does not exist yet.

It's interesting that git is workflows as defined in the PR git ref, versus what is actually in `master`.

.github/scripts/check-header.py
```diff
@@ -0,0 +1,148 @@
...
+        This is particularly useful when ingesting output from other programs, like git actions
+        '''
+        assert os.path.exists(path_json), 'No such file: ' + path_json
+
+        # Get list of files to check from JSON file
+        with open(path_json) as file_json:
+            file_checklist = load(file_json)
+            assert isinstance(file_checklist, list), 'Expected list for singular JSON List entry'
+
+        # Accrue how how files fail the check
```
cobusve, line 30: how how

.github/scripts/check-header.py
```diff
@@ -0,0 +1,148 @@
...
+            else:
+                print('File Delta: %s' % path)
+                print(*unified_diff(lines[:len(self.header)], self.header))
+                return False
+
+
+def configArgParser():
+    parser = ArgumentParser(description='FreeRTOS file header checker. We expect a consistent header across all '
+                                        'first party files. The header includes current version number, copyright, '
+                                        'and FreeRTOS license.')
```
cobusve, line 56: Are we also checking the 2 links?

cobusve (approved):

.github/scripts/check-header.py
dachalco: Yes, for now the expected `freertos_header` and `kernel_header` are set in main and include version number, copyright, license, and the two links. If anything diffs here, it reports a diff similar looking to gitdiff. Additionally any diffs in any of the checked files will fail the git action

.github/scripts/check-header.py
dachalco: @dachalco does not compute

dachalco: do you mean how is loading? If so, at top it `from json import load`.
If the JSON formatting of the loaded file is in question, perhaps I should add more detail to the assert message.
Currently the JSON intakes exclusively exists for cooperation with:
https://github.com/lots0logs/gh-action-get-changed-files

PR #419, opened 2020-11-23
Replace exponential_backoff with submodule to FreeRTOS/backoffAlgorithm

aggarw13:
A new repository, `FreeRTOS/backoffAlgorithm`, has been created for hosting the library for backoff calculation. This repo replaces the `FreeRTOS-Plus/Source/Utilities/exponential_backoff` with the submodule to the new repository, and updates all the demos that use retry logic to use the backoffAlgorithm API

PR #420, opened 2020-11-23
Fix issues with CI header check with .gitmodules and directories

aggarw13:
The PR #419 updates `.gitmodules` and adds a submodule (a.k.a a folder), but the `FreeRTOS-Header-Checker / File Header Checks` check is failing as it does not correctly handle `.gitmodules` and folders. This PR updates the script to ignore these 2 kinds of entries from being inspected

### DIFF
--- a/.github/scripts/check-header.py
+++ b/.github/scripts/check-header.py
@@ -43,6 +43,11 @@ class HeaderChecker:
         if self.isIgnoredFile(path):
             return True
 
+        # Skip if entry is a directory.
+        if os.path.isdir(path):
+            print('Skipping valid file check on directory path: %s' % path)
+            return True
+
         # Don't need entire file. Read sufficienly large chunk of file that should contain the header
         with open(path, encoding='utf-8', errors='ignore') as file:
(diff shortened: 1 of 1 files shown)


.github/scripts/check-header.py
```diff
@@ -164,7 +169,8 @@ def main():
     checker.ignoreExtension('.vcxproj',
```
lundinc2, line 169: It would make more sense to me if we only checked C files, and then manually ignore directories and files that are not to be updated with our header.

lundinc2 (approved):

.github/scripts/check-header.py
lundinc2: I acknowledge this is not really within the scope of the PR, but should you want to I think that would be a good change.

n9wxu (approved):

.github/scripts/check-header.py
aggarw13: I agree with the suggestion. Tagging @dachalco to inform about the improvement suggestion.

PR #431, opened 2020-11-30
Update incorrect file headers and variable name

yourslab:
Description
-----------

Wrong variable name was used as it is not actually a path but rather, a pointer to a char array containing the X509 certificate.

Also there are a bunch of file headers that are outdated so I've also fixed them as part of this PR.


archigup (approved):

archigup (approved):

yourslab:
Did a little experiment and found out that if you rebase with your changes already in master (from a more recent PR), then the PR just closes automatically lol

PR #437, opened 2020-12-02
Header Checker: Misc Enhancements 

dachalco:
- Avoid checking deleted files from changelist. Specifically check {added, modified, renamed}.
- Add regex pattern for exclusion
- Update exclusion list per each repo. See `FREERTOS_IGNORED_*` AND `KERNEL_IGNORED_*` in the PR file changes.

### DIFF
--- a/.github/scripts/check-header.py
+++ b/.github/scripts/check-header.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-import os, sys
+import os, sys, re
 from argparse import ArgumentParser
 from difflib import unified_diff
 from json import load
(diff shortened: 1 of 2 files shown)


.github/scripts/check-header.py
```diff
@@ -105,76 +113,352 @@ def configArgParser():
                         help    = 'Treat arguments json files that store a list of files to check.')
     return parser
 
+#--------------------------------------------------------------------------------------------------
+#                                            CONFIG  
+#--------------------------------------------------------------------------------------------------
+FREERTOS_IGNORED_EXTENSIONS = [
+    '.1',
+    '.ASM',
+    '.C',
```
cobusve, line 122: Is this right? I think cpp files should have the same header

.github/scripts/check-header.py
```diff
@@ -105,76 +113,352 @@ def configArgParser():
...
+    '.DSW',
+    '.G_C',
+    '.H',
```
cobusve, line 125: Same here

.github/scripts/check-header.py
```diff
@@ -105,76 +113,352 @@ def configArgParser():
...
+    '.bts',
+    '.ccxml',
+    '.cdkproj',
+    '.cdkws',
+    '.cfg',
+    '.cgp',
+    '.cmake',
+    '.cmd',
+    '.config',
+    '.cpp',
```
cobusve, line 158: and here

cobusve (approved):

.github/scripts/check-header.py
dachalco: It's correct for current state of repo in that no `FreeRTOS/FreeRTOS` files of these types include the header. These should be skipped. 

The changes in `header-checks.yml` now use the check as defined in the user PR so users can update as needed

PR #450, opened 2020-12-07
Add "Connection: close" header check to S3 download demos

sukhmanm:
In the download demos, S3 will close the connection with the client after 100 requests, presenting issues for larger files. Adding a check for the "Connection: close" header (sent by the server), and re-establishing the connection if found, will allow for the application to resume downloading.

### DIFF
--- a/FreeRTOS-Plus/Demo/coreHTTP_Windows_Simulator/HTTP_S3_Download/DemoTasks/S3DownloadHTTPExample.c
+++ b/FreeRTOS-Plus/Demo/coreHTTP_Windows_Simulator/HTTP_S3_Download/DemoTasks/S3DownloadHTTPExample.c
@@ -196,6 +196,11 @@ static size_t xServerHostLength;
  */
 static const char * pcRequestURI;
 
+/**
+ * @brief The status of the network connection.
+ */
+static BaseType_t xIsConnectionEstablished = pdFALSE;
+
 /*-----------------------------------------------------------*/
 
 /**
@@ -753,6 +758,15 @@ static BaseType_t prvDownloadS3ObjectFile( const TransportInterface_t * pxTransp
 
         if( xHTTPStatus == HTTPSuccess )
         {
+            /* Check that we received a valid status code */
+            if( xResponse.statusCode != httpexampleHTTP_STATUS_CODE_PARTIAL_CONTENT )
+            {
+                LogError( ( "Received response with unexpected status code: %d.", xResponse.statusCode ) );
+                xIsConnectionEstablished = pdFALSE;
+                xStatus = pdFAIL;
+                break;
+            }
+
             /* Check for the "Connection: close" header in case the connection
              * needs to be re-established. */
             if( xResponse.respFlags == HTTP_RESPONSE_CONNECTION_CLOSE_FLAG )
@@ -767,6 +781,7 @@ static BaseType_t prvDownloadS3ObjectFile( const TransportInterface_t * pxTransp
                 if( xStatus != pdPASS )
                 {
                     LogError( ( "Could not reconnect to server. Exiting loop." ) );
+                    xIsConnectionEstablished = pdFALSE;
                     break;
                 }
             }
@@ -788,21 +803,12 @@ static BaseType_t prvDownloadS3ObjectFile( const TransportInterface_t * pxTransp
             {
                 xNumReqBytes = xFileSize - xCurByte;
             }
-
-            xStatus = ( xResponse.statusCode == httpexampleHTTP_STATUS_CODE_PARTIAL_CONTENT ) ? pdPASS : pdFAIL;
         }
         else
         {
             LogError( ( "An error occurred in downloading the file. "
                         "Failed to send HTTP GET request to %s%s: Error=%s.",
                         cServerHost, pcPath, HTTPClient_strerror( xHTTPStatus ) ) );
-        }
-
-        if( xStatus != pdPASS )
-        {
-            LogError( ( "Received an invalid response from the server "
-                        "(Status Code: %u).",
-                        xResponse.statusCode ) );
         }
     }
 
--- a/FreeRTOS-Plus/Demo/coreHTTP_Windows_Simulator/HTTP_S3_Download/DemoTasks/S3DownloadHTTPExample.c
+++ b/FreeRTOS-Plus/Demo/coreHTTP_Windows_Simulator/HTTP_S3_Download/DemoTasks/S3DownloadHTTPExample.c
@@ -302,7 +302,6 @@ static void prvHTTPDemoTask( void * pvParameters )
     /* The transport layer interface used by the HTTP Client library. */
     TransportInterface_t xTransportInterface;
     TlsTransportParams_t xTlsTransportParams = { 0 };
-    BaseType_t xIsConnectionEstablished = pdFALSE;
     /* HTTP Client library return status. */
     HTTPStatus_t xHTTPStatus = HTTPSuccess;
     UBaseType_t uxDemoRunCount = 0UL;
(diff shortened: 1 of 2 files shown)


FreeRTOS-Plus/Demo/coreHTTP_Windows_Simulator/HTTP_S3_Download_Multithreaded/DemoTasks/S3DownloadMultithreadedHTTPExample.c
```diff
@@ -1132,6 +1127,36 @@ static BaseType_t prvDownloadLoop( void )
...
+        /* Check for the "Connection: close" here (instead of the response
+         * task), so that the failed request can be re-sent */
+        else if( xDownloadRespItem.xResponse.respFlags == HTTP_RESPONSE_CONNECTION_CLOSE_FLAG )
+        {
+            LogInfo( ( "'Connection:close' header found in server response. Attempting to re-connect to the server." ) );
+            vTaskSuspend( xRequestTask );
+            vTaskSuspend( xResponseTask );
+
+            /* Disconnect and re-establish the connection. */
+            TLS_FreeRTOS_Disconnect( &xNetworkContext );
```
sarenameas, line 1140: If there are multiple threads sharing this connection, do we need to synchronize access?

sarenameas: Nevermind I see that only task uses the connection.

FreeRTOS-Plus/Demo/coreHTTP_Windows_Simulator/HTTP_S3_Download/DemoTasks/S3DownloadHTTPExample.c
```diff
@@ -750,25 +753,45 @@ static BaseType_t prvDownloadS3ObjectFile( const TransportInterface_t * pxTransp
...
+
+                /* Disconnect and re-establish the connection. */
+                TLS_FreeRTOS_Disconnect( &xNetworkContext );
+                xStatus = connectToServerWithBackoffRetries( prvConnectToServer,
+                                                             &xNetworkContext );
+
+                if( xStatus != pdPASS )
+                {
+                    LogError( ( "Could not reconnect to server. Exiting loop." ) );
+                    break;
```
sarenameas, line 770: When we break here, it looks like the demo will still pass?

FreeRTOS-Plus/Demo/coreHTTP_Windows_Simulator/HTTP_S3_Download/DemoTasks/S3DownloadHTTPExample.c
```diff
@@ -750,25 +753,45 @@ static BaseType_t prvDownloadS3ObjectFile( const TransportInterface_t * pxTransp
...
+                }
+            }
+            else
```
sarenameas, line 773: If the connection: close header is found, we still want to do the stuff in this else-case. The response will still contain the partial content.

sarenameas (dismissed):

FreeRTOS-Plus/Demo/coreHTTP_Windows_Simulator/HTTP_S3_Download/DemoTasks/S3DownloadHTTPExample.c
```diff
@@ -750,6 +753,24 @@ static BaseType_t prvDownloadS3ObjectFile( const TransportInterface_t * pxTransp
 
         if( xHTTPStatus == HTTPSuccess )
         {
+            /* Check for the "Connection: close" header in case the connection
+             * needs to be re-established. */
+            if( xResponse.respFlags == HTTP_RESPONSE_CONNECTION_CLOSE_FLAG )
```
gkwicker, line 758: Why are we assuming this is happening because of some request limit?  What if the bucket/file doesn't exist, there is a policy issue, etc?  I think you'd see similar behavior.  This seems like __a bad idea__ – so why are we doing it?

Same comment at 1134.

gkwicker (commented):
What is the point of this change?

FreeRTOS-Plus/Demo/coreHTTP_Windows_Simulator/HTTP_S3_Download/DemoTasks/S3DownloadHTTPExample.c
sarenameas: This is resolved with checking for the status code before doing anything else.

FreeRTOS-Plus/Demo/coreHTTP_Windows_Simulator/HTTP_S3_Download/DemoTasks/S3DownloadHTTPExample.c
```diff
@@ -750,6 +758,34 @@ static BaseType_t prvDownloadS3ObjectFile( const TransportInterface_t * pxTransp
...
+
+                if( xStatus != pdPASS )
+                {
+                    LogError( ( "Could not reconnect to server. Exiting loop." ) );
+                    xIsConnectionEstablished = pdFALSE;
+                    break;
+                }
+            }
+
             LogDebug( ( "Received HTTP response from %s%s...",
```
sarenameas, line 789: Can we move these debug logs of the response to before line 762?

FreeRTOS-Plus/Demo/coreHTTP_Windows_Simulator/HTTP_S3_Download_Multithreaded/DemoTasks/S3DownloadMultithreadedHTTPExample.c
```diff
@@ -1134,6 +1133,41 @@ static BaseType_t prvDownloadLoop( void )
...
+                break;
+            }
+
+            /* Check for the "Connection: close" here (instead of the response
+             * task), so that the failed request can be re-sent */
+            if( xDownloadRespItem.xResponse.respFlags == HTTP_RESPONSE_CONNECTION_CLOSE_FLAG )
+            {
+                LogInfo( ( "'Connection:close' header found in server response. Attempting to re-connect to the server." ) );
+                vTaskSuspend( xRequestTask );
+                vTaskSuspend( xResponseTask );
```
sarenameas, line 1150: Do these tasks need to be suspended? Are they trying to use the connection?

FreeRTOS-Plus/Demo/coreHTTP_Windows_Simulator/HTTP_S3_Download_Multithreaded/DemoTasks/S3DownloadMultithreadedHTTPExample.c
```diff
@@ -1134,6 +1133,41 @@ static BaseType_t prvDownloadLoop( void )
...
+                    xIsConnectionEstablished = pdFALSE;
+
+                    LogError( ( "Could not reconnect to server. Exiting task." ) );
+                    break;
+                }
+            }
+
             LogInfo( ( "The HTTP task received a response from the server. Adding to response queue." ) );
 
             /* Add response to response queue. */
```
sarenameas, line 1173: I think it would be useful for the response task (prvResponseTask) to still get a non-206 response to print out the error message in the response body sent by S3.

FreeRTOS-Plus/Demo/coreHTTP_Windows_Simulator/HTTP_S3_Download_Multithreaded/DemoTasks/S3DownloadMultithreadedHTTPExample.c
sukhmanm: No, they don't use the connection, but I found while I was testing larger files that the request task would time out waiting for the queue to empty up, so I figured it would be safest to suspend until the connection was re-established.

sarenameas: Ok

sarenameas (approved):

gkwicker (commented):
I still have questions on this one:

1.  Where is this 100 request limit documented?
1. If you add a 500ms delay in the download loop, does the connection stay up?  If so, the demo is just getting throttled and you need to add a delay in the download loop.
1. If you want to modify this demo to download a larger file, could you just increase the block size?
1. If there is a request limit (and we need to provide a link to the documentation for this limit), why is it a problem that the demo can only download files up to a certain size?

Given that this is a demo application, I don't agree with this change since it makes the code more complicated.  Let's first make sure we understand the root cause why the connection is being closed.

abhidixi11:
Closing this PR for now, since not required for the release. We should reopen if needed.

PR #452, opened 2020-12-07
Map errors from FreeRTOS+TCP to mbedTLS port and interpret retriable errors

yourslab:
Description
-----------

From the FreeRTOS documentation, pdFREERTOS_ERRNO_ENOSPC means that timeout occurred before any data could be sent or received.
- In the plaintext transport-interface implementation, we would directly return `-pdFREERTOS_ERRNO_ENOSPC`. However, an error like this can occur when the TCP buffer is full, so this ought to be retriable. Libraries that consume the transport interface interpret a return value of 0 to mean that send/recv can be invoked again to get the data. As such, we should appropriately set the return value as 0 when the status is `-pdFREERTOS_ERRNO_ENOSPC`.
- In the mbedTLS port, we would directly return whatever `FreeRTOS_send` or `FreeRTOS_recv` returns. However, sometimes, the return value can be an error. In such cases, we ought to map an error from FreeRTOS+TCP to an equivalent error in mbedTLS. In the case of `-pdFREERTOS_ERRNO_ENOSPC`, we map that to `MBEDTLS_ERR_SSL_TIMEOUT`. When the mbedTLS transport-interface send/recv wrapper sees that value, it appropriately returns 0, so that the library can retry the send/recv. I've verified that when the mbedTLS port returns an error, that same error is returned by `mbedtls_ssl_write` & `mbedtls_ssl_read`.

### DIFF
--- a/FreeRTOS-Plus/Source/Application-Protocols/network_transport/freertos_plus_tcp/using_plaintext/using_plaintext.c
+++ b/FreeRTOS-Plus/Source/Application-Protocols/network_transport/freertos_plus_tcp/using_plaintext/using_plaintext.c
@@ -122,14 +122,21 @@ int32_t Plaintext_FreeRTOS_recv( NetworkContext_t * pNetworkContext,
 {
     PlaintextTransportParams_t * pPlaintextTransportParams = NULL;
     int32_t socketStatus = 0;
+    BaseType_t recvCount = 0;
 
     configASSERT( ( pNetworkContext != NULL ) && ( pNetworkContext->pParams != NULL ) );
 
     pPlaintextTransportParams = pNetworkContext->pParams;
-    socketStatus = FreeRTOS_recv( pPlaintextTransportParams->tcpSocket,
-                                  pBuffer,
-                                  bytesToRecv,
-                                  0 );
+
+    if( ( bytesToRecv > 1 ) ||
+        ( recvCount = FreeRTOS_recvcount( pNetworkContext->tcpSocket ) > 0 ) )
+    {
+        socketStatus = FreeRTOS_recv( pNetworkContext->tcpSocket, pBuffer, bytesToRecv, 0 );
+    }
+    else
+    {
+        socketStatus = ( int32_t ) recvCount;
+    }
 
     return socketStatus;
 }
--- a/FreeRTOS-Plus/Source/Application-Protocols/network_transport/freertos_plus_tcp/using_plaintext/using_plaintext.c
+++ b/FreeRTOS-Plus/Source/Application-Protocols/network_transport/freertos_plus_tcp/using_plaintext/using_plaintext.c
@@ -128,6 +128,8 @@ int32_t Plaintext_FreeRTOS_recv( NetworkContext_t * pNetworkContext,
 
     pPlaintextTransportParams = pNetworkContext->pParams;
 
+    /* When 1 byte is requested without any available data in the TCP socket's
+     * Rx stream, this function immediately returns 0. */
     if( ( bytesToRecv > 1 ) ||
         ( recvCount = FreeRTOS_recvcount( pPlaintextTransportParams->tcpSocket ) > 0 ) )
     {
--- a/FreeRTOS-Plus/Source/Application-Protocols/network_transport/freertos_plus_tcp/using_plaintext/using_plaintext.c
+++ b/FreeRTOS-Plus/Source/Application-Protocols/network_transport/freertos_plus_tcp/using_plaintext/using_plaintext.c
@@ -127,8 +127,15 @@ int32_t Plaintext_FreeRTOS_recv( NetworkContext_t * pNetworkContext,
 
     pPlaintextTransportParams = pNetworkContext->pParams;
 
-    /* When 1 byte is requested without any available data in the TCP socket's
-     * Rx stream, this function immediately returns 0 without blocking. */
+    /* The TCP socket may have a receive block time.  If bytesToRecv is greater 
+     * than 1 then a frame is likely already part way through reception and 
+     * blocking to wait for the desired number of bytes to be available is the
+     * most efficient thing to do.  If bytesToRecv is 1 then this may be a 
+     * speculative call to read to find the start of a new frame, in which case 
+     * blocking is not desirable as it could block an entire protocol agent 
+     * task for the duration of the read block time and therefore negatively 
+     * impact performance.  So if bytesToRecv is 1 then don't call recv unless 
+     * it is known that bytes are already available. */
     if( bytesToRecv == 1 )
     {
         socketStatus = ( int32_t ) FreeRTOS_recvcount( pPlaintextTransportParams->tcpSocket );
--- a/FreeRTOS-Plus/Source/Application-Protocols/network_transport/freertos_plus_tcp/using_plaintext/using_plaintext.c
+++ b/FreeRTOS-Plus/Source/Application-Protocols/network_transport/freertos_plus_tcp/using_plaintext/using_plaintext.c
@@ -121,26 +121,25 @@ int32_t Plaintext_FreeRTOS_recv( NetworkContext_t * pNetworkContext,
                                  size_t bytesToRecv )
 {
     PlaintextTransportParams_t * pPlaintextTransportParams = NULL;
-    int32_t socketStatus = 0;
-    BaseType_t recvCount = 0;
+    int32_t socketStatus = 1;
 
     configASSERT( ( pNetworkContext != NULL ) && ( pNetworkContext->pParams != NULL ) );
 
     pPlaintextTransportParams = pNetworkContext->pParams;
 
     /* When 1 byte is requested without any available data in the TCP socket's
-     * Rx stream, this function immediately returns 0. */
-    if( ( bytesToRecv > 1 ) ||
-        ( recvCount = FreeRTOS_recvcount( pPlaintextTransportParams->tcpSocket ) > 0 ) )
+     * Rx stream, this function immediately returns 0 without blocking. */
+    if( bytesToRecv == 1 )
+    {
+        socketStatus = ( int32_t ) FreeRTOS_recvcount( pPlaintextTransportParams->tcpSocket );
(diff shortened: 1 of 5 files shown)


FreeRTOS-Plus/Source/Application-Protocols/network_transport/freertos_plus_tcp/using_plaintext/using_plaintext.c
```diff
@@ -131,6 +131,14 @@ int32_t Plaintext_FreeRTOS_recv( NetworkContext_t * pNetworkContext,
                                   bytesToRecv,
                                   0 );
 
+    if( socketStatus == -pdFREERTOS_ERRNO_ENOSPC )
```
RichardBarry, line 134: I don' t think it is possible for FreeRTOS_recv() to return -pdFREERTOS_ERRNO_ENOSPC - if there are no bytes available it returns 0 already.

FreeRTOS-Plus/Source/Application-Protocols/network_transport/freertos_plus_tcp/using_plaintext/using_plaintext.c
```diff
@@ -131,6 +131,14 @@ int32_t Plaintext_FreeRTOS_recv( NetworkContext_t * pNetworkContext,
                                   bytesToRecv,
```
RichardBarry, line 131: We also want to prevent FreeRTOS_recv() from blocking when it is called to see if there is any data available, rather than when it is called because it is know there is a packet being received.  If the number of bytes requested is 1, then it is being called to look for the start of another frame, so we can prevent it blocking when no frame is available by adding:

```
    if( ( bytesToRecv > 1 ) || ( FreeRTOS_recvcount( pNetworkContext->tcpSocket ) > 0 ) )
    {
        socketStatus = FreeRTOS_recv( pNetworkContext->tcpSocket, pBuffer, bytesToRecv, 0 );
    }
    else
    {
        socketStatus = 0;
    }
```

FreeRTOS-Plus/Source/Utilities/mbedtls_freertos/mbedtls_freertos_port.c
```diff
@@ -121,13 +152,37 @@ int mbedtls_platform_recv( void * ctx,
...
+        /* No data could be sent because there was insufficient memory. */
+        case -pdFREERTOS_ERRNO_ENOMEM:
+        /* No data could be sent because xSocket was not a valid TCP socket. */
+        case -pdFREERTOS_ERRNO_EINVAL:
+            returnStatus = MBEDTLS_ERR_SSL_INTERNAL_ERROR;
+            break;
+
+        /* A timeout occurred before any data could be sent. */
+        case -pdFREERTOS_ERRNO_ENOSPC:
+            returnStatus = MBEDTLS_ERR_SSL_TIMEOUT;
```
RichardBarry, line 177: Same comment as per the plain text code - is this value even possible?

FreeRTOS-Plus/Source/Utilities/mbedtls_freertos/mbedtls_freertos_port.c
```diff
@@ -121,13 +152,37 @@ int mbedtls_platform_recv( void * ctx,
...
+            break;
+
+        default:
+            returnStatus = ( int ) recvStatus;
```
RichardBarry, line 181: What about the case where recvStatus is 0, which I think mbedTLS will take as an error, but might just mean you need to call FreeRTOS_recv() again when data is available?  I'm not sure what will happen here and the consequence might be different depending on whether the read is speculative to find the start of a new frame (in which case if there is no frame we don't want to block and just move on), compared to when a frame is known to be being received (in which case we want to read again).

RichardBarry (changes requested):

FreeRTOS-Plus/Source/Application-Protocols/network_transport/freertos_plus_tcp/using_plaintext/using_plaintext.c
yourslab: Oh right. Good point.

FreeRTOS-Plus/Source/Utilities/mbedtls_freertos/mbedtls_freertos_port.c
yourslab: Can we just return `MBEDTLS_ERR_SSL_TIMEOUT` when recvStatus == 0? mbedTLS treats 0 as a special case where the connection is closed. However, if we return `MBEDTLS_ERR_SSL_TIMEOUT` instead, then mbedTLS will treat that as a retriable error.

FreeRTOS-Plus/Source/Application-Protocols/network_transport/freertos_plus_tcp/using_plaintext/using_plaintext.c
yourslab: That's a nice addition. One thing I noticed is that `FreeRTOS_recvcount` may return `-pdFREERTOS_ERRNO_EINVAL` for an invalid socket, and we don't necessarily want that to be retriable and return 0. I've added a check for this edge case.

FreeRTOS-Plus/Source/Utilities/mbedtls_freertos/mbedtls_freertos_port.c
yourslab: I think if we call `FreeRTOS_recv` again until there is data to receive, the timeout initially passed to `TLS_FreeRTOS_Connect` function no longer applies. After diving into mbedTLS's usage of the callback, it does not seem like it will attempt any speculative reads. It always tries to receive as many bytes as it can from the stream and into the message buffer.

Another advantage with returning `MBEDTLS_ERR_SSL_TIMEOUT` is that it will attempt renegotiation with the server if it is pending.

FreeRTOS-Plus/Source/Application-Protocols/network_transport/freertos_plus_tcp/using_plaintext/using_plaintext.c
```diff
@@ -122,14 +122,24 @@ int32_t Plaintext_FreeRTOS_recv( NetworkContext_t * pNetworkContext,
...
 
     configASSERT( ( pNetworkContext != NULL ) && ( pNetworkContext->pParams != NULL ) );
 
     pPlaintextTransportParams = pNetworkContext->pParams;
-    socketStatus = FreeRTOS_recv( pPlaintextTransportParams->tcpSocket,
-                                  pBuffer,
-                                  bytesToRecv,
-                                  0 );
+
+    if( ( bytesToRecv > 1 ) ||
```
sarenameas, line 131: Is this supposed to be >= to 1?

yourslab: I think the intention here is that if `bytesToRecv` is 1, it is counted as a speculative read, so we want to check if there are bytes in the stream that are yet to be read. Otherwise, we just want to block for the entire duration of the timeout.

sarenameas: Ok, so when coreMQTT goes to read a single byte for the packet type, if there is data FreeRTOS_recvcount will be greater than zero.
But if there is no data, then zero will be returned. In this case, do we not still want to try to read from the socket for the socket's recv timeout?

yourslab: It deviates from the original behavior, but I believe Richard made this suggestion so that the receive block time doesn't put the task to sleep when an attempt is made to read a single byte.

sarenameas: Ok, so I guess we shall see if there are more nodataavailable failures in the demo after this change.

sarenameas (dismissed):

FreeRTOS-Plus/Source/Application-Protocols/network_transport/freertos_plus_tcp/using_plaintext/using_plaintext.c
yourslab: Sounds good. I ran a custom job and looks like everything still passed in CI.

aggarw13: I'm concerned with the special logic of the _speculative read_. Why make a special case only for a single byte? There is no requirement of this from the `transport_interface.h` specification. 
We don't have this behavior for the POSIX implementation or the Secure Sockets based implementation. So a special case just for FreeRTOS+TCP implementation does not sound good.

yourslab: This is suggested by @RichardBarry in an earlier comment.

aggarw13: Can you please add a comment in the code explaining the rationale? 
Also, it would be good to mention this in the API documentation for the function about the single read not waiting for the socket timeout period so that the calling code can be aware of utilizing this feature of the implementation

abhidixi11 (approved):

n9wxu (approved):

FreeRTOS-Plus/Source/Application-Protocols/network_transport/freertos_plus_tcp/using_plaintext/using_plaintext.c
RichardBarry: I updated the comment in the code.

PR #456, opened 2020-12-08
Update kernel submodule pointer to version 47338393f

cobusve:


n9wxu (approved):

FreeRTOS/Test/CBMC/patches/0009-Remove-overflow-asserts-from-xQueueGenericCreate.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/FreeRTOS/Source/queue.c b/FreeRTOS/Source/queue.c
+index b01dfd11f..b219b599a 100644
+--- a/FreeRTOS/Source/queue.c
++++ b/FreeRTOS/Source/queue.c
+@@ -395,7 +395,7 @@ BaseType_t xQueueGenericReset( QueueHandle_t xQueue,
+         xQueueSizeInBytes = ( size_t ) ( uxQueueLength * uxItemSize ); /*lint !e961 MISRA exception as the casts are only redundant for some ports. */
+ 
+         /* Check for multiplication overflow. */
+-        configASSERT( ( uxItemSize == 0 ) || ( uxQueueLength == ( xQueueSizeInBytes / uxItemSize ) ) );
++        /* configASSERT( ( uxItemSize == 0 ) || ( uxQueueLength == ( xQueueSizeInBytes / uxItemSize ) ) ); */
```
gkwicker, line 10: Why do we have commented-out code in this patch?

gkwicker (changes requested):
One quick question, see comments.

FreeRTOS/Test/CBMC/patches/0009-Remove-overflow-asserts-from-xQueueGenericCreate.patch
cobusve: The patch is removing the assert because it breaks CBMC, it is applied before the proofs run

gkwicker (approved):

PR #459, opened 2020-12-09
Moved definition of ucBufferStorage out of if macro. 

scholzp:
Description
-----------

Compilation of the StreamBufferDemo will fail currently if configSUPPORT_STATIC_ALLOCATION was set to 0 or left undefined (because it will default to 0, as mentioned in the documentation). 
This is because the variable ucBufferStorage is only defined, if configSUPPORT_STATIC_ALLOCATION is set to 1 (see line 119 of StreamBufferDemo.c) but will be used without any further checks in line 232 and in some other places in StreamBufferDemo.c.

Test Steps
-----------

Compile StreamBufferDemo without defining configSUPPORT_STATIC_ALLOCATION. For me the error occurred when compiling the "RISC-V_Renode_Emulator_SoftConsole" demo (configSUPPORT_STATIC_ALLOCATION not defined in FreeRTOSConfig.h)  with Release FreeRTOSv202011.00.

Related Issue
-----------


cobusve (approved):

cobusve:
Thanks for submitting this we have merged it.

PR #475, opened 2020-12-14
Use updated stack size configuration for HTTP multi-threaded demo

sukhmanm:
The stack size config used for task creation was initially taken from the `amazon-freertos` demo, and was causing runtime issues on Microchip. This PR reverts the change and uses `democonfigDEMO_STACKSIZE` instead, to account for platform differences as well as evolving differences in demo design. The change has been verified in CI.


FreeRTOS-Plus/Demo/coreHTTP_Windows_Simulator/HTTP_S3_Download_Multithreaded/DemoTasks/S3DownloadMultithreadedHTTPExample.c
```diff
@@ -577,14 +572,14 @@ static void prvHTTPDemoTask( void * pvParameters )
             /* Open request and response tasks. */
             xDemoStatus = xTaskCreate( prvRequestTask,
                                        "RequestTask",
-                                       httpexampleTASK_STACK_SIZE,
+                                       democonfigDEMO_STACKSIZE,
```
leegeth, line 575: What was the runtime issue that got fixed with this change?

cobusve (approved):

FreeRTOS-Plus/Demo/coreHTTP_Windows_Simulator/HTTP_S3_Download_Multithreaded/DemoTasks/S3DownloadMultithreadedHTTPExample.c
leegeth: Verified on the hardware that the demos were failing due to an allocation failure and the reduction in stack size helped fix the issue.

leegeth (approved):

PR #482, opened 2020-12-17
AutoRelease + Header Checker Upgrades

dachalco:
Description
-----------

- add `requirements.txt` for easy installation of dependencies `pip -r requirements.txt`
    - strictly depend on python libs so `requirements.txt` can be sufficient
        - Replaced the Git API to use gitpython for local edits/commits etc.
        - Replaced zipping utility with python lib
- refactored/improved some of the older scripts ('versioning.py', 'release.py')
- Add mechanism for undoing an entire flow
- git org should be required input
- Add mechanism for baselining on certain commit, and creating detached HEAD tags
- Version number updates need to update header check ref for kernel and fr/fr
    - decoupled kernel and core configs so they live in their respective repos. Need for next bullet
        - See https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/236 
- Git auth is more flexible for `release.py`. Needed to run in git action environment
- Created git action that wraps script
- Faster
- Added fix to header checker

Test Steps
-----------


.github/workflows/core-checks.yml
```diff
@@ -40,6 +40,7 @@ jobs:
       - name: Check File Headers
         run: |
           cd inspect
-          ../tools/.github/scripts/check-header.py --json ${HOME}/files_modified.json ${HOME}/files_added.json ${HOME}/files_renamed.json
+          export PYTHONPATH=tools/.github/scripts:${PYTHONPATH}
+          .github/scripts/core_checker.py --json ${HOME}/files_modified.json ${HOME}/files_added.json ${HOME}/files_renamed.json
```
lundinc2, line 44: Could we add a print statement for the file list ? May be useful to see what the script is run on

.github/scripts/common/header_checker.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python3
```
lundinc2, line 1: Would be nice to add a README for these scripts, to provide some documentation for contributors explaining the errors and how to fix them. Could link to it in the error log

lundinc2: Can you add the minimum python3 version needed as a comment or in the readme?

.github/scripts/common/header_checker.py
```diff
@@ -0,0 +1,121 @@
...
+
+def dprint(msg):
+    print('[DEBUG]: %s' % str(msg))
+
+class HeaderChecker:
+    def __init__(self, header, padding=1000, ignored_files=[], ignored_ext=[], ignored_patterns=[]):
+        self.padding = padding
+        self.header = header
+
+        self.ignorePatternList = ignored_patterns.copy()
```
lundinc2, line 16: Is this copy necessary? Why not reference the original list?

.github/scripts/core_checker.py
```diff
@@ -0,0 +1,309 @@
...
+]
+
+FREERTOS_IGNORED_FILES = [
+    'fyi-another-way-to-ignore-file.txt',
+    'mbedtls_config.h'
+]
+
+FREERTOS_HEADER = [
+    '/*\n',
+    ' * FreeRTOS V202011.00\n',
```
lundinc2, line 267: Should this be V202012.00

.github/scripts/common/header_checker.py
dachalco: It's a quirk of python, namely optional function args have static refs. If instead I did ` = ignored_patterns`, then every `instance.ignorePatternList` would ref to the same list. The copy creates a new ref that is unique to the object and not shared

.github/scripts/core_checker.py
dachalco: It should! Thank you!

.github/scripts/common/header_checker.py
dachalco: These are both good points adding now

n9wxu (approved):

.github/workflows/core-checks.yml
dachalco: Added this

.github/scripts/common/header_checker.py
dachalco: Added a descriptive help message that instructs on how to fix the failing check and added a comment to indicate python version

.github/scripts/common/header_checker.py
lundinc2: Maybe instead of an empty list, they could be set to None and checked against, would the copy be needed then? This is more for my understanding, this code doesn't really need to be optimized, but i have not yet looked to see how long this check runs for

.github/scripts/common/header_checker.py
```diff
@@ -0,0 +1,135 @@
...
+        # Get list of files to check from JSON file
+        with open(path_json) as file_json:
+            file_checklist = load(file_json)
+            assert isinstance(file_checklist, list), 'Expected list for singular JSON List entry'
+
+        # Accrue how how files fail the check
+        n_failed = 0
+        for path_file in file_checklist:
+            assert isinstance(path_file, str), 'Unexpected JSON format for ' + path_json
+            if os.path.exists(path_file):
```
lundinc2, line 35:
```suggestion
            if os.path.exists(path_file) and not self.isValidFile(path_file):
                n_failed += 1
```

.github/scripts/versioning.py
```diff
@@ -69,27 +69,25 @@ def ask_yes_no_question(question):
...
+            continue
 
         for f in files:
-            if ext_filter != None:
-                ext = '.' + f.split('.')[-1]
-                if ext in ext_filter:
+            if exclude_hidden and f[0] == '.':
+                continue
+
+            if (ext_filter is None
```
lundinc2, line 88: May be worth wrapping the AND expression in a () to emphasis the precedence of the logical operators

.github/scripts/release.py
```diff
@@ -86,60 +110,148 @@ def printReleases(self):
...
+        # Remove from remote
+        if self.tag in self.local_repo.tags:
+            info('Deleting tag "%s"' % self.tag)
+            self.local_repo.remote('origin').push(':%s' % self.tag)
+        else:
+            info('A tag does not exists for "%s". No need to delete.' % self.tag)
+
+    def updateSubmodulePointer(self, rel_path, ref):
+        submodule = Repo(rel_path)
+        submodule.remote('origin').fetch()
```
lundinc2, line 137: We want to use the latest submodule tags, this could lead to a commit on a development branch I think

.github/scripts/release.py
```diff
@@ -86,60 +110,148 @@ def printReleases(self):
...
+        if self.tag not in self.local_repo.tags:
+            error('Could not find a SHA to rollback to for tag "%s"' % self.tag)
+            return False
+
+        # Search for auto release SHAs that match the release tag SHA
+        tag_commit = self.local_repo.tag('refs/tags/%s' % self.tag).commit
+        prior_commit = self.local_repo.commit(tag_commit.hexsha + '~%d' % n_autocommits)
+        n_commits_searched = 0
+        for commit in self.local_repo.iter_commits():
+            if n_commits_searched > 25:
```
lundinc2, line 181: Why 25?

.github/scripts/release.py
```diff
@@ -148,96 +260,140 @@ def createGitRelease(self):
...
-        self.zip = create_package(zip_root_path,
-                                  rel_repo_path,
-                                  'FreeRTOSv%s' % self.version,
-                                  exclude_files=FREERTOS_RELATIVE_FILE_EXCLUDES)
+        if os.path.exists(rel_repo_path):
+            shutil.rmtree(rel_repo_path)
+
+        # Download a fresh copy for packaging
+        info('Downloading fresh copy of %s for packing...' % zip_name, end='')
+        packaged_repo = Repo.clone_from(self.getRemoteEndpoint(self.repo_name),
```
lundinc2, line 361: --depth 1 will decrease the amount of stuff cloned, could be worth investigating

.github/workflows/auto-release.yml
```diff
@@ -0,0 +1,52 @@
+name: Auto-Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      commit_id:
+        description: 'Commit ID'
+        required: true
+        default: 'HEAD'
```
lundinc2, line 9: I think it would be better to force a specific commit ID to be specified

.github/workflows/auto-release.yml
```diff
@@ -0,0 +1,52 @@
...
+      version_number:
+        description: 'Version Number (Ex. 202000.00)'
+        required: true
+        default: '202000.00'
```
lundinc2, line 13: Same here, having a default seems like it could trip things up, may as well force input

dachalco: Absolutely

.github/workflows/auto-release.yml
dachalco: At the moment most of the releases baseline onto HEAD. But need to keep option available for the much less frequent non-HEAD baselined releases

.github/scripts/release.py
dachalco: It's arbitrary, but would fit better as an optional arg. Adding now

dachalco: It's arbitrary, refactoring to at least make it an optional arg

dachalco: done

n9wxu (approved):

PR #502, opened 2021-02-12
Unit Test/timer[0]

dachalco:
Description
-----------
Initial iteration. Only includes happy path test for `xTimerCreate`. Realized I really wanted an integration test which I'll be writing elsewhere, but made this along the way.

Also make CC/LD assignment conditional and only set in Makefile if not already set, so users like myself can specify CC/LD without updating the Makefile.

Test Steps
-----------
Read `FreeRTOS/Test/Cmock/ReadMe.md`.
```
make timers run
```

Related Issue
-----------

### DIFF
--- a/FreeRTOS/Test/CMock/timers/timers_utest.c
+++ b/FreeRTOS/Test/CMock/timers/timers_utest.c
@@ -1,0 +1,123 @@
+/*
+ * FreeRTOS V202012.00
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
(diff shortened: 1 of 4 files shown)


FreeRTOS/Test/CMock/timers/timers_utest.c
```diff
@@ -0,0 +1,123 @@
...
+#include "mock_queue.h"
+#include "mock_list.h"
+
+
+/* ============================  GLOBAL VARIABLES =========================== */
+static uint16_t usMallocFreeCalls = 0;
+
+/* ==========================  CALLBACK FUNCTIONS =========================== */
+
+void * pvPortMalloc( size_t xSize )
```
lundinc2, line 48: These callback functions seem like they should be incrementing / decrementing `usMallocFreeCalls`

n9wxu (approved):

n9wxu (approved):

PR #509, opened 2021-02-16
QEMU CORTEX_M3_MPS2_QEMU_GCC Demo Failure Counter

lundinc2:
Add failure counters to QEMU CORTEX_M3_MPS2_QEMU_GCC Full Demo and GDB script to count failures.

Description
-----------
Add failure counters to QEMU CORTEX_M3_MPS2_QEMU_GCC Full Demo and GDB script to count failures. The updated variables store how many times a task has faltered, and a GDB script is provided that can be used to extract the failures when QEMU is running with no serial.

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/FreeRTOS/Demo/CORTEX_M3_MPS2_QEMU_GCC/main_full.c
+++ b/FreeRTOS/Demo/CORTEX_M3_MPS2_QEMU_GCC/main_full.c
@@ -189,6 +189,40 @@ static char *pcStatusMessage = "OK: No errors";
 semaphore tracing API functions.  It has no other purpose. */
 static SemaphoreHandle_t xMutexToDelete = NULL;
 
+/* Counters for task failures. */
+long long unsigned ullSuccesfulIterations = 0;
+#if( configUSE_PREEMPTION != 0 )
+long long unsigned ullTimerDemoTaskFailures = 0;
+#endif
+long long unsigned ullStreamBufferTaskFailures = 0;
+long long unsigned ullMessageBufferFailures = 0;
+long long unsigned ullTaskNotificationFailures = 0;
+long long unsigned ullInterruptSemaphoreFailures = 0;
+long long unsigned ullEventGroupTaskFailures = 0;
+long long unsigned ullIntegerMathsFailures = 0;
+long long unsigned ullGenericQueueFailures = 0;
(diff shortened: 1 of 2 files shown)


RichardBarry:
This is a good idea, but I don't think it will work for a couple of reasons.  First errors discovered in tests are latched - so once a test reports an error it will continue to do so every time the test is re-checked.  Second the check task uses "else if" for each check, so only the first failure on each will result in the counter getting incremented.

You may also like to check "main_full.c" in this project to ensure the GCC project is running all the same tests: https://github.com/FreeRTOS/FreeRTOS/pull/503

FreeRTOS/Demo/CORTEX_M3_MPS2_QEMU_GCC/main_full.c
```diff
@@ -285,21 +319,25 @@ HeapStats_t xHeapStats;
 			/* These tasks are only created when preemption is used. */
 			if( xAreTimerDemoTasksStillRunning( xCycleFrequency ) != pdTRUE )
 			{
+				ullTimerDemoTaskFailures++;
 				pcStatusMessage = "Error: TimerDemo";
 			}
 		}
 		#endif
 
 		if( xAreStreamBufferTasksStillRunning() != pdTRUE )
```
RichardBarry, line 328: Added this into the conversation - should have added here:

This is a good idea, but I don't think it will work for a couple of reasons. First errors discovered in tests are latched - so once a test reports an error it will continue to do so every time the test is re-checked. Second the check task uses "else if" for each check, so only the first failure on each will result in the counter getting incremented.

You may also like to check "main_full.c" in this project to ensure the GCC project is running all the same tests: #503

RichardBarry (changes requested):

FreeRTOS/Demo/CORTEX_M3_MPS2_QEMU_GCC/main_full.c
lundinc2: I will go through that link and check. I think it would still be valuable to keep counting, even though it is latched, so we can approximate how early it fell over? I'll change the "else if" statements to if, and create a separate boolean to see how many cycles it was running successfully for.

PR #513, opened 2021-02-20
Add tests that catch a race condition in the reported size of a message buffer

RichardBarry:
Description
-----------
Introduce tasks that test the coherency of the reported space available in a message buffer from two separate tasks.  Designed to highlight the issue reported in https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/264

Introduce configRUN_ADDITIONAL_TESTS which must be set to 1 to run the new tests.  That is because the new tests got added to an existing standard demo file and smaller platforms may not have the resources to run them.

Set configRUN_ADDITIONAL_TESTS to 1 in the MSVC and IAR/QEMU project so both project run the new test.  Also add missing 'volatile' qualifier on some register accesses in the same file.

Test Steps
-----------
Run the soak test in either the MSVC Win32 or the IAR MPS2 QEMU projects.

Related Issue
-----------

https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/264


RichardBarry:
PR is not supposed to include the Xilinx changes hence it was recreated by Cobus.

PR #528, opened 2021-03-17
Adding unit tests for stream buffer and message buffer

ravibhagavandas:
Description
-----------

* Added CMock unit tests for stream buffer and message buffer
* Coverage  - 100% line, branch and functional coverage
* Update kernel submodule pointer to latest main to get the fix for stream buffer failing test.

Test Steps
-----------
Run only stream buffer tests:
`make stream_buffer`

Run only message buffer tests:
`make message_buffer`

Run whole suite with coverage:
`make`


alfred2g (dismissed):

FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
```diff
@@ -0,0 +1,670 @@
...
+}
+
+/**
+ * @brief Validates message buffer creation through a static buffer allocated by the user.
+ */
+void test_xMessageBufferCreateStatic_success( void )
+{
+    StaticMessageBuffer_t messageBufferStruct = { 0 };
+    /* The size of message buffer array should be one greater than the required size of message buffer. */
+    uint8_t messageBufferArray[ TEST_MESSAGE_BUFFER_SIZE + 1 ] = { 0 };
```
paulbartell, line 332: It would be safer to heap allocate your static buffers so that the address sanitizer can check correct access.

FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
```diff
@@ -0,0 +1,670 @@
+/*
```
paulbartell, line 1: Please add @ coverage tags to this file so that the lcov / lcovhtml targets will work correctly.

FreeRTOS/Test/CMock/stream_buffer/stream_buffer_utest.c
```diff
@@ -0,0 +1,1107 @@
+/*
```
paulbartell, line 1: Please add @ coverage tags to this file so that the lcov / lcovhtml targets will work correctly.

paulbartell (changes requested):

FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
```diff
@@ -0,0 +1,670 @@
...
+/* ==============================  Test Cases  ============================== */
+
+/**
+ * @brief Validates that message buffer of sample test size is created successfully.
+ */
+void test_xMessageBufferCreate_success( void )
+{
+    xMessageBuffer = xMessageBufferCreate( TEST_MESSAGE_BUFFER_SIZE );
+    TEST_ASSERT_NOT_EQUAL( NULL, xMessageBuffer );
+    validate_message_buffer_init_state( xMessageBuffer, TEST_MESSAGE_BUFFER_SIZE );
```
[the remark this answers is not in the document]
ravibhagavandas: Size of the memory allocated for stream buffer depends on the internal structure (implementation) used and is not directly exposed from API. I think checking the internal size would not add much value here as we are testing the API. However we do check for memory leaks from tests.

FreeRTOS/Test/CMock/stream_buffer/stream_buffer_utest.c
```diff
@@ -0,0 +1,1107 @@
...
+}
+
+/* ==============================  Test Cases  ============================== */
+
+/**
+ * @brief Validates that stream buffer of sample size is created succesfully.
+ */
+void test_xStreamBufferCreate_success( void )
+{
+    xStreamBuffer = xStreamBufferCreate( TEST_STREAM_BUFFER_SIZE, TEST_STREAM_BUFFER_TRIGGER_LEVEL );
```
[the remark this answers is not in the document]
ravibhagavandas: Size of the memory allocated for stream buffer depends on the internal structure (implementation) used and is not directly exposed from API.

FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
ravibhagavandas: As per docs, address sanitizer should catch buffer overflow on stack memory as well.
https://github.com/google/sanitizers/wiki/AddressSanitizerExampleStackOutOfBounds

paulbartell: Yes. My bad. I confirmed that AddressSanitizer handles stack allocated variables as well on GCC 9.3.0.

FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
ravibhagavandas: This is fixed in coverage filter script. So not required.

FreeRTOS/Test/CMock/stream_buffer/stream_buffer_utest.c
ravibhagavandas: This is fixed in coverage filter script. So not required.

FreeRTOS/Test/CMock/stream_buffer/stream_buffer_utest.c
```diff
@@ -0,0 +1,1107 @@
...
+    TEST_ASSERT_EQUAL( NULL, xStreamBuffer );
+
+    /* Assertion fails if a zero buffer size is passed as  the parameter. */
+    if( TEST_PROTECT() )
+    {
+        ( void ) xStreamBufferCreate( 0, TEST_STREAM_BUFFER_TRIGGER_LEVEL );
+    }
+
+    validate_and_clear_assertitions();
```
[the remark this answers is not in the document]
ravibhagavandas: We need to fail the test if assertion fails, hence not included in TEST_PROTECT

FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
```diff
@@ -0,0 +1,670 @@
...
+    TEST_ASSERT_NULL( xMessageBuffer );
+    validate_and_clear_assertitions();
+
+    /* Returns NULL when NULL message buffer struct is passed as a parameter. */
+    xMessageBuffer = xMessageBufferCreateStatic( sizeof( messageBufferArray ), messageBufferArray, NULL );
+    TEST_ASSERT_NULL( xMessageBuffer );
+    validate_and_clear_assertitions();
+
+    /* Should only assert if the size passed is less than the minimum size requried to store metadata size. */
+    if( TEST_PROTECT() )
```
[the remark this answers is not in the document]
ravibhagavandas: TEST_PROTECT() here is to catch abort from source code on assertion and not to continue the test case. These test cases are simple parameter validations so grouped them together to one test.

FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
```diff
@@ -0,0 +1,670 @@
...
+/*******************************************************************************
+ * Unity fixtures
+ ******************************************************************************/
+void setUp( void )
+{
+    assertionFailed = 0;
+    xMessageBuffer = NULL;
+    senderTaskWoken = 0;
+    receiverTaskWoken = 0;
+    shouldAbortOnAssertion = pdTRUE;
```
paulbartell, line 236: Please call the _Init function for each of your mocked files. Otherwise, the test cases are tightly coupled and failures could cascade.
```
mock_task_Init();
mock_fake_assert_Init();
mock_fake_port_Init();
```

FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
```diff
@@ -0,0 +1,670 @@
...
+    vFakeAssert_StubWithCallback( vFakeAssertStub );
+    /* Track calls to malloc / free */
+    UnityMalloc_StartTest();
+}
+
+/*! called before each testcase */
+void tearDown( void )
+{
+    TEST_ASSERT_EQUAL_MESSAGE( 0, assertionFailed, "Assertion check failed in code." );
+    UnityMalloc_EndTest();
```
paulbartell, line 251: Please call the _Verify() and _Destory() functions for each mocked file:

```
mock_task_Verify();
mock_task_Destroy();

mock_fake_assert_Verify();
mock_fake_assert_Destroy();

mock_fake_port_Verify();
mock_fake_port_Destroy();
```

FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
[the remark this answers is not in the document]
paulbartell: It's worth checking this against the size of StaticStreamBuffer_t.

FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
```diff
@@ -0,0 +1,670 @@
...
+    xMessageBuffer = xMessageBufferCreate( TEST_MESSAGE_BUFFER_SIZE );
+    TEST_ASSERT_NOT_NULL( xMessageBuffer );
+
+    /* Should assert if a null data buffer is passed. */
+    if( TEST_PROTECT() )
+    {
+        ( void ) xMessageBufferReceive( xMessageBuffer, NULL, TEST_MAX_MESSAGE_SIZE, TEST_MESSAGE_BUFFER_WAIT_TICKS );
+    }
+
+    validate_and_clear_assertitions();
```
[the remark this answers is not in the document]
paulbartell: The problem is that this will cause the test to fail because the memory allocated by xMessageBufferCreate will never be freed by the call to vStreamBufferDelete.

FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
```diff
@@ -0,0 +1,739 @@
...
+
+    /* Should assert if a null message buffer handle is passed. */
+    if( TEST_PROTECT() )
+    {
+        ( void ) xMessageBufferReceive( NULL, message, TEST_MAX_MESSAGE_SIZE, TEST_MESSAGE_BUFFER_WAIT_TICKS );
+    }
+
+    validate_and_clear_assertitions();
+
+    vStreamBufferDelete( xMessageBuffer );
```
paulbartell, line 623: Similar issue here re: memory allocation / de-allocation.

paulbartell: I think it's not the end of the world to leak a single xMessageBuffer though, especially since this is only running in a simulated environment.

paulbartell (approved):
I suggest verifying that the amount of memory allocated by the xStreamBufferGenericCreate function is equal to the size of StreamBufferHandle_t.

Otherwise, this test is acceptable to me.

paulbartell (dismissed):

paulbartell:
You're missing coverage of some of the assert cases. Check the report at this link: https://github.com/FreeRTOS/FreeRTOS/suites/2346659470/artifacts/49575098, which is generated with the make lcovhtml target.

paulbartell (approved):

cobusve (approved):

PR #545, opened 2021-04-02
Fix infinite loop due to TEST_PROTECT usage in message buffer and stream buffer unit tests

paulbartell:
Fix infinite loop due to TEST_PROTECT usage in message buffer and stream buffer unit tests

Description
-----------
Replace TEST_PROTECT usage with a try/catch/throw using the CException library.

Issue repro steps
-----------
To reproduce the problem, remove a configASSERT that is expected by one of the two unit tests and note that the relevant test case enters an infinite loop.

### DIFF
--- a/FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
+++ b/FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
@@ -340,7 +340,7 @@ void test_xMessageBufferCreate_integer_overflow( void )
 }
 
 /**
- * @breif API should fail and return null if the malloc fails.
+ * @brief API should fail and return null if the malloc fails.
  */
 void test_xMessageBufferCreate_malloc_fail( void )
 {
--- a/FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
+++ b/FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
@@ -29,7 +29,7 @@
 #include <stdlib.h>
 #include <stdbool.h>
 
-/* Queue includes */
+/* Message Buffer includes */
 #include "FreeRTOS.h"
 #include "FreeRTOSConfig.h"
 #include "message_buffer.h"
@@ -37,6 +37,7 @@
 /* Test includes. */
 #include "unity.h"
 #include "unity_memory.h"
+#include "CException.h"
 
 /* Mock includes. */
 #include "mock_task.h"
@@ -71,6 +72,31 @@
  * below trigger level while receiveing data.
  */
 #define TEST_MESSAGE_BUFFER_WAIT_TICKS       ( 1000U )
+
+/**
+ * @brief CException code for when a configASSERT should be intercepted.
+ */
+#define configASSERT_E                       0xAA101
+
+/**
+ * @brief Expect a configASSERT from the funciton called.
+ *  Break out of the called function when this occurs.
+ * @details Use this macro when the call passsed in as a parameter is expected
+ * to cause invalid memory access.
+ */
+#define EXPECT_ASSERT_BREAK( call )             \
+    do                                          \
+    {                                           \
+        shouldAbortOnAssertion = true;          \
+        CEXCEPTION_T e = CEXCEPTION_NONE;       \
+        Try                                     \
+        {                                       \
+            call;                               \
+            TEST_FAIL();                        \
+        }                                       \
+        Catch( e )                              \
+        TEST_ASSERT_EQUAL( configASSERT_E, e ); \
+    } while ( 0 )
 
 /* ============================  GLOBAL VARIABLES =========================== */
 
@@ -131,7 +157,7 @@ static void vFakeAssertStub( bool x,
 
         if( shouldAbortOnAssertion == pdTRUE )
         {
-            TEST_ABORT();
+            Throw( configASSERT_E );
         }
     }
 }
@@ -328,10 +354,7 @@ void test_xMessageBufferCreate_malloc_fail( void )
  */
 void test_xMessageBufferCreate_zero_size( void )
 {
-    if( TEST_PROTECT() )
-    {
-        ( void ) xMessageBufferCreate( 0 );
-    }
+    EXPECT_ASSERT_BREAK( ( void ) xMessageBufferCreate( 0 ) );
 
     validate_and_clear_assertions();
 }
@@ -341,10 +364,7 @@ void test_xMessageBufferCreate_zero_size( void )
  */
 void test_xMessageBufferCreate_invalid_size( void )
 {
-    if( TEST_PROTECT() )
-    {
-        ( void ) xMessageBufferCreate( TEST_MESSAGE_METADATA_SIZE );
-    }
+    EXPECT_ASSERT_BREAK( ( void ) xMessageBufferCreate( TEST_MESSAGE_METADATA_SIZE ) );
 
     validate_and_clear_assertions();
 }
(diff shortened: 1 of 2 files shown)


FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
```diff
@@ -328,25 +361,31 @@ void test_xMessageBufferCreate_malloc_fail( void )
  */
 void test_xMessageBufferCreate_zero_size( void )
 {
-    if( TEST_PROTECT() )
-    {
-        ( void ) xMessageBufferCreate( 0 );
-    }
+    xMessageBuffer = xMessageBufferCreate( 0 );
```
ravibhagavandas, line 364: This should test for an assertion from the code [here](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/main/stream_buffer.c#L248) and the test should not continue after that.

FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
```diff
@@ -328,25 +361,31 @@ void test_xMessageBufferCreate_malloc_fail( void )
...
 /**
  * @brief Should assert if the size passed is less than the minimum size requried to store metadata size.
  */
 void test_xMessageBufferCreate_invalid_size( void )
 {
-    if( TEST_PROTECT() )
-    {
-        ( void ) xMessageBufferCreate( TEST_MESSAGE_METADATA_SIZE );
-    }
+    xMessageBuffer = xMessageBufferCreate( TEST_MESSAGE_METADATA_SIZE );
```
ravibhagavandas, line 380: Same here. Test should not continue after the assertion is hit.

FreeRTOS/Test/CMock/stream_buffer/stream_buffer_utest.c
```diff
@@ -455,40 +488,44 @@ void test_xStreamBufferCreate_malloc_fail( void )
...
 void test_xStreamBufferCreate_zero_buffer_size( void )
 {
+    xStreamBuffer = xStreamBufferCreate( 0, TEST_STREAM_BUFFER_TRIGGER_LEVEL );
 
-    if( TEST_PROTECT() )
-    {
-        ( void ) xStreamBufferCreate( 0, TEST_STREAM_BUFFER_TRIGGER_LEVEL );
-    }
-    validate_and_clear_assertions();
+    validate_and_clear_multiple_assertions();
```
ravibhagavandas, line 497: This test case specifically checks if assertion is called and program is aborted on passing a zero size. I think we shouldn't continue the program execution and check for other assertions. Here, the test case is trying to do multiple things.

FreeRTOS/Test/CMock/stream_buffer/stream_buffer_utest.c
```diff
@@ -455,40 +488,44 @@ void test_xStreamBufferCreate_malloc_fail( void )
...
  * @brief Assertion fails if trigger level is greater than the stream buffer size.
  */
 void test_xStreamBufferCreate_invalid_trigger_level( void )
 {
-    if( TEST_PROTECT() )
-    {
-        ( void ) xStreamBufferCreate( TEST_STREAM_BUFFER_SIZE, ( TEST_STREAM_BUFFER_SIZE + 1 ) );
-    }
-    validate_and_clear_assertions();
+    xStreamBuffer = xStreamBufferCreate( TEST_STREAM_BUFFER_SIZE, ( TEST_STREAM_BUFFER_SIZE + 1 ) );
```
ravibhagavandas, line 510: Why is the assertion turned off and we continue execution here.

FreeRTOS/Test/CMock/stream_buffer/stream_buffer_utest.c
```diff
@@ -455,40 +488,44 @@ void test_xStreamBufferCreate_malloc_fail( void )
...
  * @brief Assertion fails while trying to delete a null stream buffer.
  */
 void test_xStreamBufferDelete_null_stream_buffer( void )
 {
-    if( TEST_PROTECT() )
-    {
-        vStreamBufferDelete( NULL );
-    }
-    validate_and_clear_assertions();
+    EXPECT_ASSERT_BREAK( vStreamBufferDelete( NULL ) );
```
ravibhagavandas, line 526: I think this could also be solved by adding a `TEST_ABORT()` after `vStreamBufferDelete()` which is similar to what is done in macro, but uses unity provided macros.

FreeRTOS/Test/CMock/stream_buffer/stream_buffer_utest.c
```diff
@@ -554,12 +588,13 @@ void test_xStreamBufferCreateStatic_invalid_trigger_level( void )
     /* The size of stream buffer array should be one greater than the required size of stream buffer. */
     uint8_t streamBufferArray[ TEST_STREAM_BUFFER_SIZE + 1 ] = { 0 };
 
-    if( TEST_PROTECT() )
-    {
-        ( void ) xStreamBufferCreateStatic( sizeof( streamBufferArray ), TEST_STREAM_BUFFER_SIZE + 2, streamBufferArray, &streamBufferStruct );
-    }
+    xStreamBuffer = xStreamBufferCreateStatic( sizeof( streamBufferArray ), TEST_STREAM_BUFFER_SIZE + 2, streamBufferArray, &streamBufferStruct );
```
ravibhagavandas, line 591: Test case shouldn't continue here after assert.

FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
paulbartell: Asserts can be turned off in production. This tests both the configASSERT disabled or configASSERT enabled.

FreeRTOS/Test/CMock/message_buffer/message_buffer_utest.c
paulbartell: Asserts can be turned off in production. This tests both the configASSERT disabled and configASSERT enabled cases.

FreeRTOS/Test/CMock/stream_buffer/stream_buffer_utest.c
paulbartell: Asserts can be turned off in production. This tests both the configASSERT disabled and configASSERT enabled cases.

FreeRTOS/Test/CMock/stream_buffer/stream_buffer_utest.c
paulbartell: Asserts can be turned off in production. This tests both the configASSERT disabled and configASSERT enabled cases.

FreeRTOS/Test/CMock/stream_buffer/stream_buffer_utest.c
paulbartell: I don't think you understand the issue.

In your original code, If any assertion in validate_and_clear_assertions() fails, execution flow returns to line 487... So it enters an endless loop.

ravibhagavandas: Thanks for the clarification. The issue is with TEST_ABORT() internally called from within `TEST_ASSERT_EQUAL()` in `validate_and_clear_assertions()`

ravibhagavandas (dismissed):

dan4thewin (approved):

PR #553, opened 2021-04-08
Add test for backlogged auto-reload timer

jefftenney:
#### Add Test for backlogged auto-reload timer

Description
-----------
Test to verify an application request to stop an auto-reload timer is handled correctly when the auto-reload timer is backlogged.  When the reload action for an auto-reload timer finds the new timer period has already elapsed, the timer is backlogged.

Test Steps
-----------
Choose a full demo to run, strip it down to just the timer demo, add a call to `vTimerDemoIncludeBacklogTests( pdTRUE )`, and run the demo.  (https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/303 fixes an issue with backlogged auto-reload timers, and this new test triggers the issue.)

Related Issue
-----------
https://forums.freertos.org/t/freertos-time-correction/11856
https://forums.freertos.org/t/timer-start-dont-trace-race-condition/12168

### DIFF
--- a/FreeRTOS/Demo/Common/Minimal/TimerDemo.c
+++ b/FreeRTOS/Demo/Common/Minimal/TimerDemo.c
@@ -696,18 +696,15 @@ static void prvTest7_CheckBacklogBehaviour( void )
 		configASSERT( xTestStatus );
 	}
 
-	ucAutoReloadTimerCounters[ 0 ] = ( uint8_t ) 0;
-
 	/* Prompt the callback function to stop the timer. */
 	ucIsStopNeededInTimerZeroCallback = ( uint8_t ) pdTRUE;
 
 	/* Now start the timer.  This will appear to happen immediately to
 	this task because this task is running at a priority below the timer
-	service task.  Use a timer period small enough that the call to
-	xTaskCatchUpTicks() below has a small impact on other tests that might
-	be running, but large enough that other tests won't cause an extra
-	(unexpected) timer expiry. */
-	#define tmrdemoBACKLOG_TIMER_PERIOD		( ( TickType_t ) 5 )
+	service task.  Use a timer period of one tick so the call to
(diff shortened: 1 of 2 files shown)


n9wxu (dismissed):

cobusve (approved):

FreeRTOS/Demo/Common/Minimal/TimerDemo.c
```diff
@@ -664,6 +684,75 @@ uint8_t ucTimer;
...
+		xTestStatus = pdFAIL;
+		configASSERT( xTestStatus );
+	}
+
+	/* Verify the backlog cleared properly. */
+	if( ucAutoReloadTimerCounters[ 0 ] != ( uint8_t ) tmrdemoEXPECTED_BACKLOG_EXPIRIES )
+	{
+		xTestStatus = pdFAIL;
+		configASSERT( xTestStatus );
+	}
```
jefftenney, line 738: On second thought, before we enshrine this check as a requirement and have some user start depending on it, maybe we shouldn't.

What's the harm if the timer task processes the stop command and decides not to work through the entire backlog?  Callers already accept ambiguity in *when* timer commands are processed.  So perhaps it's better to detect an error here with `>` instead of `!=`, verifying only that the stop command didn't get processed too late.  

Or even better eliminate this check altogether.  That would allow us to reduce `tmrdemoBACKLOG_TIMER_PERIOD` to 1 and make this test less disruptive to other tests because the value passed to `xTaskCatchUpTicks()` would be very small (2).

Either way a change here will slightly simplify and reduce https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/305.

jefftenney:
@n9wxu @cobusve Apologies for round 2 here, but I'm actually done now.  Wanted to withdraw one of the checks in the new test.

n9wxu (approved):

PR #554, opened 2021-04-08
Small RISC-V spike demo improvements

timsifive:
Small RISC-V spike demo improvements

Description
-----------
* Put XLEN into .o files. This makes it easier to work on RV32 and RV64 binaries side-by-side.
* Let the debugger disable HTIF use by making its use conditional on a global volatile int.
* Makefile now links the binary at BASE_ADDRESS. I need this so I can easily generate the appropriate binaries for riscv-tests/debug. Unfortunately there doesn't seem to be any good mechanism to externally define values for lds files, so I'm running it through the C preprocessor.

Test Steps
-----------
Tested RV32 and RV64 against spike as part of riscv-tests/debug. (Those changes have not yet been submitted.)


n9wxu (approved):

timsifive:
Thanks for responding so quickly!

PR #566, opened 2021-04-17
Revert "Add test for timer start delayed past expiration"

RichardBarry:
Reverts FreeRTOS/FreeRTOS#557 as it is negatively impacting the stream buffer tests even when running on the most powerful targets.  We can investigate and adjust as necessary, then re-merge.


jefftenney:
For whatever it's worth, I thought about how to make the timer backlog test less disruptive but nothing came to mind.  Those calls to `xTaskCatchUpTicks()` are just asking for trouble with any concurrent testing going on.

Let's put this in TimerDemo.h:

image

PR #568, opened 2021-04-20
Automate doxygen ZIP creation with GitHub Action

aggarw13:
Automate creation of doxygen ZIP output by using the custom `doxygen` GitHub Action, that has been updated to generate ZIP artifact in [PR](https://github.com/FreeRTOS/CI-CD-Github-Actions/pull/21).


paulbartell (approved):

.github/workflows/ci.yml
```diff
@@ -22,3 +24,37 @@ jobs:
...
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+          path: freertos
+      - name: Install Python3
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.7.10'
```
paulbartell, line 38: If you use a python-version of "3.7" or "3.8" github will automatically use the latest released version.

.github/workflows/ci.yml
```diff
@@ -22,3 +24,37 @@ jobs:
...
+        with:
+          path: ./freertos
+          # List of directories containing libraries whose doxygen output will be generated.
+          libs_parent_dir_path: FreeRTOS-Plus/Source,FreeRTOS-Plus/Source/AWS,FreeRTOS-Plus/Source/Application-Protocols
+          generate_zip: true
+      - name: Upload doxygen artifact if main branch
+        if: success() && github.ref == 'refs/heads/main'
+        uses: actions/upload-artifact@v2
+        with:
+          name: doxygen.zip
```
paulbartell, line 57: I would prefer that there not be a zip file within a zip file here.

.github/workflows/ci.yml
```diff
@@ -22,3 +24,37 @@ jobs:
...
+          path: ./doxygen.zip
+          retention-days: 2
```
paulbartell, line 59: Should we retain the artifacts for a longer time?

.github/workflows/ci.yml
aggarw13: The file is uploaded only as a single ZIP file, but GitHub only exports every artifact as an extra ZIP.

.github/workflows/ci.yml
aggarw13: Good question, I don't have any preference on number of days for retention. We have used 2 days as the retention period in [C-SDK's ci.yml](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/main/.github/workflows/ci.yml#L262), and am keeping the same period.

.github/workflows/ci.yml
aggarw13: Sure, changed

.github/workflows/ci.yml
paulbartell: This isn't really ideal. The output from the github action could really use some cleanup as well.

A few suggestions:
* Add a git short revision hash to the end of the directory that gets zipped
* Get rid of the intermediate "Source" directory inside FreeRTOS-Plus
* Collapse the docs/doxygen/output/html directories in each module.
* Should coreJSON and corePKCS11 be inside the Application-Protocols directory?

.github/workflows/ci.yml
paulbartell: What's the intended use of the artifacts? Do we plan to generate the public documentation from this github action?

.github/workflows/ci.yml
```diff
@@ -22,3 +24,37 @@ jobs:
...
+          path: freertos
+      - name: Install Python3
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+      - name: Download tag dependency of coreMQTT-Agent
+        run: |
+          # We don't need to generate the coreMQTT docs, we only need the tag file.
+          # Therefore, we can just download it.
+          mkdir -p freertos/FreeRTOS-Plus/Source/Application-Protocols/coreMQTT-Agent/source/dependency/coreMQTT/docs/doxygen/output
```
paulbartell, line 43: I would prefer that we regenerate the documentation for coreMQTT from whatever revision the sub module is currently at, since the docs on FreeRTOS.org do not appear to be versioned.

.github/workflows/ci.yml
paulbartell: I guess it's fine, since we could just re-run the job if we need to regenerate the documentation. We need to use the specific revision of each repository to make this build reproducible.

.github/workflows/ci.yml
aggarw13: The action is a generic utility which can be used from both C-SDK and FreeRTOS hub repositories. The folder structure does not affect the documentation as the libraries are individually parsed by the documentation before hosting on doxygen. 
At this point, updating the action for reaching the ideal goal may not be suitable. We can definitely consider these as improvements for the future. 

As part of the artifact uploading, I can add the git hash within the ZIP name.

.github/workflows/ci.yml
aggarw13: >What's the intended use of the artifacts? Do we plan to generate the public documentation from this github action?

Yes

paulbartell (approved):
Approving this so that we don't block the release.

.github/workflows/ci.yml
aggarw13: That is certainly the right/ideal way to do it, however, currently, the coreMQTT-Agent has a direct dependency on the hosted version of coreMQTT doxygen (as can be seen [here](https://github.com/FreeRTOS/coreMQTT-Agent/blob/134b478a4862aa084575e3c37ab9fc86c04d3509/docs/doxygen/config.doxyfile#L2229)) and thus, it is required for the `mqtt.tag` to match the hosted documentation on freertos.org.

In the future, when the coreMQTT submodule is advanced past the LTS tag, we will have to find a better way to address the dependency.

dan4thewin (approved):

PR #585, opened 2021-04-29
Merge FreeRTOS 202104.00 to main

aggarw13:
Merge [FreeRTOS 202104.00](https://github.com/FreeRTOS/FreeRTOS/tree/202104.00) release to `main`.


dachalco (approved):
There's a bunch of files failing the header check. We can skip them for now, and let them resolve in other PRs, rather than doing it all at once here.

aggarg (approved):

PR #587, opened 2021-05-03
Fix versioning of header checks

dachalco:
Description
-----------
Adjusts regex pattern so it doesn't replace formatting in the header checks.

Test Steps
-----------
Run test release on your fork, verify that `core-checker.py` has appropriate versioning done.

`python3 release.py <your-git-username> --new-core-version=<test-version> --use-git-ssh --core-repo-branch=main`

`python3 release.py <your-git-username> --new-kernel-version=<test-version> --use-git-ssh --core-repo-branch=main`

Related Issue
-----------
n/a

### DIFF
--- a/.github/scripts/versioning.py
+++ b/.github/scripts/versioning.py
@@ -104,9 +104,9 @@ def extract_version_number_from_file(file_path):
         match = re.search('\s*\*\s*(Amazon FreeRTOS.*V(.*))', content, re.MULTILINE)
         # Is it a kernel file?
         if match is None:
-            match = re.search('\s*\*\s*(FreeRTOS Kernel.*V(.*))', content, re.MULTILINE)
+            match = re.search('\s*\*\s*(FreeRTOS Kernel.*V([0-9]*\.[0-9]*\.[0-9]*))', content, re.MULTILINE)
         if match is None:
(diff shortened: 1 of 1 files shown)


.github/scripts/versioning.py
```diff
@@ -104,9 +104,9 @@ def extract_version_number_from_file(file_path):
         match = re.search('\s*\*\s*(Amazon FreeRTOS.*V(.*))', content, re.MULTILINE)
         # Is it a kernel file?
         if match is None:
-            match = re.search('\s*\*\s*(FreeRTOS Kernel.*V(.*))', content, re.MULTILINE)
+            match = re.search('\s*\*\s*(FreeRTOS Kernel.*V([0-9]*\.[0-9]*\.[0-9]*))', content, re.MULTILINE)
```
lundinc2, line 107: Not a blocker, but would be great to paste comments with each expression that provide an example string the regex is looking for. In this case it is fairly obvious, it just tends to make maintenance a bit easier

lundinc2, line 107: Not a blocker, but would be great to paste comments with each expression that provide an example string the regex is looking for. In this case it is fairly obvious, it just tends to make maintenance a bit easier

lundinc2 (approved):

gkwicker (approved):

PR #592, opened 2021-05-08
Test: Mask unit test warning

alfred2g:
Mask unit test warning

Description
-----------
Jenkins was failing the unit test because of a warning, removing the warning as it is previously tested

Test Steps
-----------
Unit Tests pass


FreeRTOS/Test/CMock/queue/generic/queue_create_dynamic_utest.c
```diff
@@ -38,6 +38,7 @@
 #include "queue.h"
 
 /* ============================  GLOBAL VARIABLES =========================== */
+extern bool xMaskAssertAndAbort;
```
paulbartell, line 41: I'm not sure I understand why this needs to be moved around.

paulbartell: Get/Set functions would be cleaner if declaring this in the header file is problematic.

alfred2g: we can't declare variables inside header files with multiple translation units
the linker will be confused with multiple declarations.
Yeah +1 for Set/Get functions

paulbartell (approved):

PR #603, opened 2021-05-19
Remove or rework assumptions in queue proofs

dan4thewin:
This commit is paired with another to queue.c in the kernel.  To accommodate significant performance changes in newer versions of CBMC, the --pointer-overflow-check flag is removed.  The --pointer-check and --bounds-check flags remain to verify pointers upon dereference.  The configASSERT() macro is changed to evaluate to nothing allowing CBMC to more fully check the code, rather than the prior practice of adding assertions to evade asserts.

These proofs all pass without errors, and run in a small amount of time.


FreeRTOS/Test/CBMC/proofs/Queue/QueueGenericCreate/QueueGenericCreate_harness.c
```diff
@@ -32,19 +32,14 @@
...
+void harness()
+{
+    UBaseType_t uxQueueLength;
+    UBaseType_t uxItemSize;
+    uint8_t ucQueueType;
 
-	size_t uxQueueStorageSize;
-	__CPROVER_assume(uxQueueStorageSize < (UINT32_MAX>>8));
+    /* Allow CBMC to run in a reasonable amount of time. */
+    __CPROVER_assume( ( uxQueueLength == QUEUE_LENGTH ) || ( uxItemSize == QUEUE_ITEM_SIZE ) );
```
AniruddhaKanhere, line 42: This assumption seems very restrictive...

If there is no other way, can we at least define the `QUEUE_LENGTH` and `QUEUE_ITEM_SIZE` to various values in the configurations file above? That would broaden the scope of the proofs.

FreeRTOS/Test/CBMC/proofs/Queue/QueueGenericCreateStatic/QueueGenericCreateStatic_harness.c
```diff
@@ -31,32 +31,22 @@
...
-
     uint8_t ucQueueType;
+    size_t storageSize;
 
-    __CPROVER_assume(uxQueueStorageSize < (UINT32_MAX>>8));
-
-    // QueueGenericReset does not check for multiplication overflow
-    __CPROVER_assume(uxItemSize < uxQueueStorageSize/uxQueueLength);
+    /* Allow CBMC to run in a reasonable amount of time. */
+    __CPROVER_assume( ( uxQueueLength == QUEUE_LENGTH ) || ( uxItemSize == QUEUE_ITEM_SIZE ) );
```
AniruddhaKanhere, line 42: Similar comment as above.

AniruddhaKanhere (changes requested):

FreeRTOS/Test/CBMC/proofs/Queue/QueueGenericCreate/QueueGenericCreate_harness.c
dan4thewin: It's binding one or the other, but not both at once; i.e., when queue length is fixed, item size can be anything, and when item size is fixed, queue length can be anything.  I used macros with the intent of providing additional values in a future PR.

AniruddhaKanhere: Makes sense. Thanks.

markrtuttle (approved):

AniruddhaKanhere (approved):

aggarg (approved):

dan4thewin:
Thanks to Kareem and Oscar, we have evidence that the proofs CI checks are passing.  I'll merge this PR now.

PR #614, opened 2021-05-27
Update unit tests to match changes in queue.c

dan4thewin:
This PR updates the kernel unit tests to match the changes in https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/328 .

Now that queue creation functions return NULL for error cases that were previously only asserted upon, many test cases are no longer possible and have been removed.


AniruddhaKanhere (approved):

paulbartell:
Please look over the coverage report at https://github.com/FreeRTOS/FreeRTOS/suites/2851256611/artifacts/63657651.

Some notes: 
* xQueueGenericReset now has reduced coverage. Likely you'll need to change the assertion to not call longjmp now that the function has been updated to check parameters.
* The configASSERT( pxNewQueue ) on line 383 xQueueGenericCreateStatic cant be covered. Please add "* LCOV_EXCL_BR_LINE */" to exclude the line from branch coverage.
* Similar issue with the assertion on line 451 in xQueueGenericCreate.
* Enable configUSE_TRACE_FACILITY in each of the FreeRTOSConfig.h configs to fix the trace macro coverage (its also causing some wrong line numbers in the coverage data).
* You've removed the test cases that exercise line 740 / 775 ( uxInitialCount <= uxMaxCount ) 
* Line 3027 has an assert that is never exercised
* line 3066 has an else that is never exercised.

FreeRTOS/Test/CMock/queue/semaphore/counting_semaphore_utest.c
```diff
@@ -69,145 +69,6 @@ int suiteTearDown( int numFailures )
...
-    TEST_ASSERT_EQUAL( QUEUE_T_SIZE, getLastMallocSize() );
-
-    /* Verify that an xSemaphoreGive fails */
-    TEST_ASSERT_EQUAL( pdFALSE, xSemaphoreGive( xSemaphore ) );
-
-    vSemaphoreDelete( xSemaphore );
-}
-
-/**
- * @brief Test xSemaphoreGive with xSemaphoreCreateCounting( 1, 2 )
```
paulbartell, line 125: One of these test cases is necessary for coverage of 
```
( uxInitialCount <= uxMaxCount ) )
```
in xQueueCreateCountingSemaphore

FreeRTOS/Test/CMock/queue/sets/queue_in_set_utest.c
```diff
@@ -75,47 +75,6 @@ int suiteTearDown( int numFailures )
 /* ==============================  Test Cases =============================== */
 
 
-/**
```
paulbartell, line 78: Removing this test case might be what's causing lower coverage in prvNotifyQueueSetContainer.

paulbartell (approved):
Thanks for updating the unit tests. Please address the decrease in coverage before merging if feasible.

PR #648, opened 2021-07-15
Add note in coreSNTP demo about Network Time Security (NTS)

aggarw13:
This PR makes the following changes in the `coreSNTP` demo: 
* To suggest the most secure way of using SNTP communication, we are **adding a documentation note** about the **Network Time Security** in the SNTP demo. 
* **Hygiene update** of using the **FreeRTOS/backoffAlgorithm** utility library for calculating time polling period interval backoff when time requests are rejected by a time server.

### DIFF
--- a/FreeRTOS-Plus/Demo/coreSNTP_Windows_Simulator/DemoTasks/SNTPClientTask.c
+++ b/FreeRTOS-Plus/Demo/coreSNTP_Windows_Simulator/DemoTasks/SNTPClientTask.c
@@ -632,8 +632,11 @@ static void closeUdpSocket( Socket_t * pSocket );
  * is configured, through the democonfigLIST_OF_SERVERS macro, and the single server
  * rejects time requests.
  *
- * @param[in] firstBackoffAttempt Flag to indicate if this is first attempt backoff is being
- * calculated in a sequence of retrying server rejections of time request.
+ * @param[in, out] pContext The context representing the back-off parameters. This
+ * context is initialized by the function whenever the caller indicates it with the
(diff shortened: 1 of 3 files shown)


FreeRTOS-Plus/Demo/coreSNTP_Windows_Simulator/DemoTasks/SNTPClientTask.c
```diff
@@ -1379,6 +1431,45 @@ static void closeUdpSocket( Socket_t * pSocket )
 
 /*-----------------------------------------------------------*/
 
+static bool calculateBackoffForNextPoll( bool firstBackoffAttempt,
+                                         uint32_t minPollPeriod,
+                                         uint32_t * pPollPeriod )
+{
+    uint16_t newPollPeriod = 0U;
+    BackoffAlgorithmStatus_t status;
+    static BackoffAlgorithmContext_t backoffParams;
```
yourslab, line 1440: I feel like this does not need to be static. It is no longer safe to use if you need to calculate backoff while something else also needs to calculate backoff.

aggarw13: I was trying to encapsulate all the types of BackoffAlgorithm algorithm within this function. I agree that it makes the re-use of this function not feasible. So, I have moved the declaration of the `BackoffAlgorithmContext_t` context out and made it a parameter to the function

yourslab (approved):

paulbartell (approved):

PR #655, opened 2021-07-19
Fix clock drift issue in SNTP demo

aggarw13:
### Issue
The existing **clock discipline algorithm** in the `coreSNTP` demo yields an inaccurate wall-clock time (in UTC) that is always behind the astronomical UTC time by the chosen polling interval period of the demo. The issue with the clock discipline algorithm was that it was calculating the **slew rate** to make the system clock be gradually compensated for the previous time synchronization clock drift over the period of the subsequent polling time interval. Thereby, the clock slew rate correction calculated is always lower than it should be (because it does not incorporate possible clock drift that would occur over another period of time polling interval till the next time synchronization attempt). 

### Fix
This PR fixes the clock discipline algorithm to make the demo calculate its system time that matches the astronomical UTC time.
The fix involves simplifying the clock discipline algorithm to use a combination of **step** and **slew** correction approaches where: 
1) **step** correction is applied on EVERY time successful synchronization with a time server. This immediately corrects the system time to match server time.
2) **slew** rate is calculated ONCE (on the second successful time synchronization with server) and applied henceforth, on every time polling interval period to _compensate gradually_ for the system clock drift during the period.

### DIFF
--- a/FreeRTOS-Plus/Demo/coreSNTP_Windows_Simulator/DemoTasks/SNTPClientTask.c
+++ b/FreeRTOS-Plus/Demo/coreSNTP_Windows_Simulator/DemoTasks/SNTPClientTask.c
@@ -439,7 +439,7 @@ static void sntpClient_GetTime( SntpTimestamp_t * pCurrentTime );
  * 2. "Slew" correction approach is used for compensating system clock drift
  *    during the poll interval period between time synchronization attempts with
  *    time server(s) when latest time server is not known. The "slew rate" is
- *    calculated ONLY once on the occassion of the second successful time
+ *    calculated ONLY once on the occasion of the second successful time
  *    synchronization with a time server. This is because the demo initializes
  *    system time with (the first second of) the democonfigSYSTEM_START_YEAR
  *    configuration, and thus, the the actual system clock drift over a period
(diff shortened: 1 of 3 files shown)


FreeRTOS-Plus/Demo/coreSNTP_Windows_Simulator/DemoTasks/SNTPClientTask.c
```diff
@@ -434,30 +430,33 @@ static void sntpClient_GetTime( SntpTimestamp_t * pCurrentTime );
...
- *    the system time may have drifted behind the server time, and can be corrected
- *    gradually over the SNTP client's polling interval period.
+ * 2. "Slew" correction approach is used for compensating system clock drift
+ *    during the poll interval period between time synchronization attempts with
+ *    time server(s) when latest time server is not known. The "slew rate" is
+ *    calculated ONLY once on the occasion of the second successful time
+ *    synchronization with a time server. This is because the demo initializes
+ *    system time with (the first second of) the democonfigSYSTEM_START_YEAR
+ *    configuration, and thus, the the actual system clock drift over a period
+ *    of time ca be calculated only AFTER the demo system time has been synchronized
```
yourslab, line 446:
```suggestion
 *    of time can be calculated only AFTER the demo system time has been synchronized
```

yourslab (approved):

cobusve (approved):

PR #661, opened 2021-07-20
Revert relative URL reversion

### LINKED ISSUE #21: Use a relative URL for the FreeRTOS-Kernel submodule.
The `url` option for the FreeRTOS-Kernel submodule can just be `../FreeRTOS-Kernel.git` rather than the full github URL. This would help organizations that would like to mirror both of these repos without having the FreeRTOS-Kernel submodule always point at github, which is not desirable.

muneebahmed10:
Description
-----------

Reverts PR #27, which was a reversion of #24. Resolves #21.

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "FreeRTOS/Source"]
 	path = FreeRTOS/Source
-	url = https://github.com/FreeRTOS/FreeRTOS-Kernel.git
+	url = ../FreeRTOS-Kernel.git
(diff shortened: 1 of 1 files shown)


muneebahmed10:
I will close this PR as there are a lot more submodules now than when the issue was created, and they cannot all be converted to use relative links. Additionally, relative links make it harder to clone from forks unless the relevant submodule has been forked as well.

PR #675, opened 2021-08-07
Fix ALPN TLS connection issue with AWS IoT Core for MQTT and HTTP demos

aggarw13:
There was incorrect configuration of the ALPN protocol string in  MQTT and HTTP demos that caused problems in using TLS ALPN extension when connecting to AWS IoT Core. This PR fixes the protocol strings to allow the mutually authenticated demos to TLS connect to AWS IoT Core on port 443 with ALPN extension.


archigup (dismissed):

n9wxu (approved):

paulbartell:
This was addressed by PR #831

PR #717, opened 2021-11-17
Remove using_mbedtls folder and move its content to the parent folder

andysun2015:
Description
-----------
Move the contents of `FreeRTOS/FreeRTOS-Plus/Source/Application-Protocols/network_transport/using_mbedtls/` to its parent folder i.e. `FreeRTOS/FreeRTOS-Plus/Source/Application-Protocols/network_transport/`. This removes one unnecessary folder in the hierarchy. 


FreeRTOS-Plus/Demo/coreMQTT_Windows_Simulator/MQTT_Multitask/WIN32.vcxproj.filters
```diff
@@ -437,7 +437,8 @@
     <ClCompile Include="..\..\..\Source\Application-Protocols\network_transport\using_mbedtls\using_mbedtls\using_mbedtls.c">
```
yourslab, line 437: I notice we have two subdirectories under `using_mbedtls`:
```
using_mbedtls
..using_mbedtls
..using_mbedtls_pkcs11
```
It is not obvious what `using_mbedtls`. We can rename it to something like `using_mbedlts_no_pkcs11`. However, it might be better to just remove it out of subdirectory and put both folders in upper level directory.

andysun2015: I'd like to make it in upper layer, will make the changes.

yourslab (approved):

aggarg (approved):

PR #831, opened 2022-07-26
Fix ALPN string format to ensure compatibility with mbedtls

paulbartell:
Fix ALPN string usage to ensure compatibility with mbedtls.

### DIFF
--- a/FreeRTOS-Plus/Demo/AWS/Mqtt_Demo_Helpers/mqtt_demo_helpers.c
+++ b/FreeRTOS-Plus/Demo/AWS/Mqtt_Demo_Helpers/mqtt_demo_helpers.c
@@ -165,21 +165,6 @@
     "?SDK=" democonfigOS_NAME "&Version=" democonfigOS_VERSION \
     "&Platform=" democonfigHARDWARE_PLATFORM_NAME "&MQTTLib=" democonfigMQTT_LIB
 
-/**
- * @brief The length of the MQTT metrics string expected by AWS IoT.
- */
-#define AWS_IOT_METRICS_STRING_LENGTH    ( ( uint16_t ) ( sizeof( AWS_IOT_METRICS_STRING ) - 1 ) )
-
-/**
- * @brief ALPN (Application-Layer Protocol Negotiation) protocol name for AWS IoT MQTT.
- *
- * This will be used if democonfigMQTT_BROKER_PORT is configured as 443 for the AWS IoT MQTT broker.
- * Please see more details about the ALPN protocol for AWS IoT MQTT endpoint
- * in the link below.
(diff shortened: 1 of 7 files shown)


FreeRTOS-Plus/Demo/AWS/Mqtt_Demo_Helpers/mqtt_demo_helpers.c
```diff
@@ -178,8 +178,13 @@
  * in the link below.
  * https://aws.amazon.com/blogs/iot/mqtt-with-tls-client-authentication-on-port-443-why-it-is-useful-and-how-it-works/
  */
-#define AWS_IOT_MQTT_ALPN                "\x0ex-amzn-mqtt-ca"
+#define AWS_IOT_ALPN_MQTT_CA_AUTH        "x-amzn-mqtt-ca"
```
n9wxu, line 181: Collecting the ALPN strings in one place is a good idea.  I think the selection of the ALPN string should also be done here instead of the #ifdef later in the code (line 358).  By collecting the decision to this place, future code changes to add ALPN's can be completely made in a single place.  Additionally, checking the code to see which choice has been can be done close to the definitions of the choices.

FreeRTOS-Plus/Demo/AWS/Mqtt_Demo_Helpers/mqtt_demo_helpers.c
```diff
@@ -350,7 +355,11 @@ static TlsTransportStatus_t prvConnectToServerWithBackoffRetries( NetworkContext
     /* ALPN protocols must be a NULL-terminated list of strings. Therefore,
      * the first entry will contain the actual ALPN protocol string while the
      * second entry must remain NULL. */
-    char * pcAlpnProtocols[] = { NULL, NULL };
+    #ifdef democonfigCLIENT_USERNAME
```
n9wxu, line 358: This decision should be performed near line 187 where the ALPN choices are defined.

n9wxu (changes requested):
Please consolidate the ALPN choices with their explanations into a single location.

FreeRTOS-Plus/Demo/AWS/Mqtt_Demo_Helpers/mqtt_demo_helpers.c
paulbartell: I disagree. Customers will likely want to adjust the selected alpn based on the port number or auth mechanism at runtime.

n9wxu: I agree with that idea.  We should make the code work that way.

n9wxu (approved):

AniruddhaKanhere (approved):

PR #846, opened 2022-09-01
Add unit test for index>=0 check

AniruddhaKanhere:
Description
-----------
Corresponding Kernel PR: https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/552

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/FreeRTOS-Plus/Source/Application-Protocols/network_transport/tcp_sockets_wrapper/ports/cellular/tcp_sockets_wrapper.c
+++ b/FreeRTOS-Plus/Source/Application-Protocols/network_transport/tcp_sockets_wrapper/ports/cellular/tcp_sockets_wrapper.c
@@ -349,6 +349,10 @@ static BaseType_t prvNetworkRecvCellular( const cellularSocketWrapper_t * pCellu
     {
         retRecvLength = ( BaseType_t ) recvLength;
     }
+    else if( socketStatus == CELLULAR_SOCKET_CLOSED )
+    {
+        retRecvLength = TCP_SOCKETS_ERRNO_ECLOSED;
+    }
     else
     {
         LogError( ( "prvNetworkRecv failed %d", socketStatus ) );


FreeRTOS/Test/CMock/tasks/tasks_1_utest.c
```diff
@@ -3550,6 +3550,29 @@ void test_vTask_Set_Get_ThreadLocalStoragePointer_success( void )
     TEST_ASSERT_EQUAL_PTR( pValue, ret_pValue );
 }
 
+/* this test ensures that if the index is negative, then NULL is returned. */
+void test_vTask_Set_Get_ThreadLocalStoragePointer_successNegative( void )
+{
+    TaskHandle_t task_handle;
+    void * ret_pValue;
+    int i = 0xABCD;
+    void * value = &i;
```
courage, line 3559: should be pValue?

FreeRTOS/Test/CMock/tasks/tasks_1_utest.c
```diff
@@ -3594,6 +3617,33 @@ void test_vTask_Set_ThreadLocalStoragePointer_fail( void )
...
+void test_vTask_Set_ThreadLocalStoragePointer_failNegative( void )
+{
+    TaskHandle_t task_handle;
+    uint32_t i = 454545;
+    void * pValue = &i;
+
+    /* Setup */
+    create_task_priority = 3;
+    task_handle = create_task();
+    ptcb = task_handle;
```
courage, line 3630: This global seems to be used throughout the file, but I don't understand why.

FreeRTOS/Test/CMock/tasks/tasks_1_utest.c
```diff
@@ -3550,6 +3550,29 @@ void test_vTask_Set_Get_ThreadLocalStoragePointer_success( void )
...
+    void * value = &i;
+
+    /* Setup */
+    create_task_priority = 3;
+    task_handle = create_task();
+    /* Expectations */
+    /* API Call */
+
+    #pragma GCC diagnostic ignored "-Warray-bounds"
+    *( task_handle->pvThreadLocalStoragePointers - 1 ) = value;
```
courage, line 3568: Not clear to me that this line is a good idea since you're going to corrupt part of the TCB. Maybe instead your verification step should be a short loop that scans backwards through enough negative indices that you know you're reading past the beginning of the TCB. The preceding fields won't all be zero, so you know you won't be getting a false negative.

Alternatively you could use xTaskCreateStatic to place the TCB at the end of a buffer that will be initially filled with a known, non-NULL, pattern.

Or just get rid of this check if we're sure the underflow will be detected by other tools.

FreeRTOS/Test/CMock/tasks/tasks_1_utest.c
```diff
@@ -3594,6 +3617,33 @@ void test_vTask_Set_ThreadLocalStoragePointer_fail( void )
...
+    task_handle = create_task();
+    ptcb = task_handle;
+    /* Expectations */
+    /* API Call */
+    vTaskSetThreadLocalStoragePointer( task_handle,
+                                       -1,
+                                       pValue );
+
+    #pragma GCC diagnostic ignored "-Warray-bounds"
+    void * value = *( ptcb->pvThreadLocalStoragePointers - 1 );
```
courage, line 3638: This operation doesn't seem as bad as the one on line 3568, but still seems a bit gross.

FreeRTOS/Test/CMock/tasks/tasks_1_utest.c
```diff
@@ -3594,6 +3617,33 @@ void test_vTask_Set_ThreadLocalStoragePointer_fail( void )
...
+    #pragma GCC diagnostic error "-Warray-bounds"
+
+    /* this wall cause a warning, since we are reading past the end of the
```
courage, line 3641: s/this wall/this will/. Does it actually cause a warning given the suppression pragmas above?

FreeRTOS/Test/CMock/tasks/tasks_1_utest.c
```diff
@@ -3550,6 +3550,29 @@ void test_vTask_Set_Get_ThreadLocalStoragePointer_success( void )
     TEST_ASSERT_EQUAL_PTR( pValue, ret_pValue );
 }
 
+/* this test ensures that if the index is negative, then NULL is returned. */
```
paulbartell, line 3553: I think it makes sense to just skip this test case when building with sanitizers enabled.

PR #862, opened 2022-10-21
coreMQTT_Windows_Simulator Update

Skptak:
coreMQTT_Windows_Simulator Update

Description
-----------
Changes to the solution to import in static versions of libraries + kernel

Test Steps
-----------
Open demo, set necessary variables, run the demo.


FreeRTOS-Plus/WindowsSimulator/FreeRTOS-Kernel/FreeRTOSConfig.h
```diff
@@ -144,7 +144,7 @@
  * results in the wired network being used, while setting
  * configNETWORK_INTERFACE_TO_USE to 2 results in the wireless network being
  * used. */
-#define configNETWORK_INTERFACE_TO_USE      ( 1L )
```
paulbartell, line 147: Perhaps this should go in the demo config?

Skptak: I'd agree with that, I'll remove it here

PR #868, opened 2022-11-03
Demo: Cortex m4 mpu mps2 with qemu

alfred2g:
Demo: Cortex m4 mpu mps2 with qemu

Description
-----------
Demo: Cortex m4 mpu mps2 with qemu

Test Steps
-----------
manual testing with qemu


FreeRTOS/Demo/CORTEX_MPU_M4_MPS2_QEMU_GCC/startup.c
```diff
@@ -0,0 +1,253 @@
...
+ * https://www.FreeRTOS.org
+ * https://github.com/FreeRTOS
+ *
+ */
+
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+//#include "CMSIS/CMSDK_CM3.h"
```
paulbartell, line 31: These look left over? Surprising that you could do this without CMSIS at all.

FreeRTOS/Demo/CORTEX_MPU_M4_MPS2_QEMU_GCC/Makefile
```diff
@@ -0,0 +1,87 @@
...
+CFLAGS                +=    -Wstrict-aliasing
+CFLAGS                +=    -Wno-error=address-of-packed-member
+CFLAGS                +=    $(INCLUDE_DIRS)
+CFLAGS                +=    -fno-builtin
+CFLAGS                +=    -ggdb3 -Og -save-temps=obj
+
+LDFLAGS                =    -T mps2_m4.ld
+LDFLAGS               +=    -Xlinker -Map=${BUILD_DIR}/output.map
+LDFLAGS               +=    -nostartfiles
+LDFLAGS               +=    -specs=/opt/gcc-arm-none-eabi-10-2020-q4-major/arm-none-eabi/lib/thumb/v7-r+fp.sp/softfp/nosys.specs
```
paulbartell, line 58:
```suggestion
LDFLAGS               +=    -specs=nosys.specs
```

FreeRTOS/Demo/CORTEX_MPU_M4_MPS2_QEMU_GCC/startup.c
```diff
@@ -0,0 +1,253 @@
...
+        " mrsne r0, psp                                                      \n"
+        " ldr r1, handler5_address_const                                      \n"
+        " bx r1                                                              \n"
+        " handler5_address_const: .word vHandleMemoryFault                    \n"
+    );
+}
+
+void Debug_Handler( void )
+{
+    __asm volatile
```
paulbartell, line 186: Not all of these are memory faults.

FreeRTOS/Demo/CORTEX_MPU_M4_MPS2_QEMU_GCC/syscall.c
```diff
@@ -0,0 +1,186 @@
...
+ */
+int _write( __attribute__( ( unused ) ) int file,
+            __attribute__( ( unused ) ) char * buf,
+            int len )
+{
+    int todo;
+
+    for( todo = 0; todo < len; todo++ )
+    {
+        UART_DR( UART0_ADDR ) = *buf++;
```
paulbartell, line 102: Semihosting is enabled, so perhaps that would be a better option for stdout?

paulbartell: Regardless, this code would likely drop random bytes on real hardware once the FIFO is full (1-byte on this platform).

paulbartell:
Looks reasonable. Can we easily target other mps2 boards with the same project?

FreeRTOS/Demo/CORTEX_MPU_M4_MPS2_QEMU_GCC/Makefile
```diff
@@ -0,0 +1,87 @@
...
+
+
+DEFINES               :=    -DQEMU_SOC_MPS2
+DEFINES               :=    -DHEAP4
+DEFINES               +=    -DportUSING_MPU_WRAPPERS=1
+
+CPPFLAGS              +=    $(DEFINES)
+
+CFLAGS                +=    -mthumb
+CFLAGS                +=    -mfloat-abi=softfp
```
paulbartell, line 41: do we want soft or hard fp? Perhaps we should build a few different variants.

alfred2g: the M3 is running with hard, 
i had some problems running this one with hard, so did it with soft

FreeRTOS/Demo/CORTEX_MPU_M4_MPS2_QEMU_GCC/startup.c
alfred2g: yeah :)

FreeRTOS/Demo/CORTEX_MPU_M4_MPS2_QEMU_GCC/syscall.c
alfred2g: doesn't really matter, as this demo does not print anything, it has to be monitored with gdb

FreeRTOS/Demo/CORTEX_MPU_M4_MPS2_QEMU_GCC/startup.c
alfred2g: nope

alfred2g: yeah

paulbartell: Rather than multiple of the same handler, can you just set each of the faults to use the same handler in the vector table?

Or build handlers that provide a bit more info upon failure for each class of fault.

FreeRTOS/Demo/CORTEX_MPU_M4_MPS2_QEMU_GCC/syscall.c
paulbartell: So perhaps this should be stubbed out then?

alfred2g: we do have a print somewhere on error, and I prefer to keep it this way in case someone copies it, as it would be complete

FreeRTOS/Demo/CORTEX_MPU_M4_MPS2_QEMU_GCC/startup.c
alfred2g: done

FreeRTOS/Demo/CORTEX_MPU_M4_MPS2_QEMU_GCC/Makefile
```diff
@@ -0,0 +1,87 @@
+CC                     =    arm-none-eabi-gcc
+BIN                   :=    RTOSDemo.axf
+
+BUILD_DIR             :=    build
+
+FREERTOS_DIR_REL      :=    ../../../FreeRTOS
+FREERTOS_DIR          :=    $(abspath $(FREERTOS_DIR_REL))
+KERNEL_DIR            :=    $(FREERTOS_DIR)/Source
+
+FREERTOS_PLUS_DIR_REL :=    ../../../FreeRTOS-Plus
```
paulbartell, line 10:
```suggestion
```

FreeRTOS/Demo/CORTEX_MPU_M4_MPS2_QEMU_GCC/Makefile
```diff
@@ -0,0 +1,87 @@
...
+FREERTOS_PLUS_DIR     :=    $(abspath $(FREERTOS_PLUS_DIR_REL))
```
paulbartell, line 11:
```suggestion
```

paulbartell (changes requested):
Semihosting usage appears to be broken here and calling printf fails. This need to be addressed prior to merging.

PR #876, opened 2022-11-18
Remove binary and generated files from Kernel Demos

paulbartell:
Update .gitignore and remove generated files from Kernel Demos

Description
-----------
Remove binary and generated files from Kernel Demos
Update .gitignore file accordingly


RichardBarry:
makefsdata.exe is an lwIP utility required to create the static file system (just C structures that get compiled into the elf file) for their webserver demo.  It's used in the currently documented Zynq XSDK project.  We can either keep that file, or preferably switch our documentation to instead point to the existing (but not yet documented) Zynq demo that builds with Vitis (the current Xilinx tool), runs in QEMU, and doesn't use lwIP (hence no makefsdata.exe file).

paulbartell:
> makefsdata.exe is an lwIP utility required to create the static file system (just C structures that get compiled into the elf file) for their webserver demo. It's used in the currently documented Zynq XSDK project. We can either keep that file, or preferably switch our documentation to instead point to the existing (but not yet documented) Zynq demo that builds with Vitis (the current Xilinx tool), runs in QEMU, and doesn't use lwIP (hence no makefsdata.exe file).

@RichardBarry: The build info for makefsdata.exe is included in the project. Unless we plan to sign the binary, it is preferable to have users build it themselves.

Regarding the FlashMB9BF506.out files, these are generated by the flashing utility, so do not need to be kept in the repository.

RichardBarry:
👍

n9wxu (approved):

AniruddhaKanhere (approved):

PR #891, opened 2022-12-02
Fix and improve WIN32-MSVC demo in FreeRTOS/Demo.

jasonpcarroll:
Description
-----------

This PR fixes the WIN32-MSVC demo in FreeRTOS/Demo and adds the following changes to functionality:
* Trace has been added to the blinky demo.
* Trace recorder continues after saving the Trace output to a file rather than just stopping altogether after a trace dump, allowing users to output / update the trace as many times as they want while the demo runs.
* Keyboard input handling has been moved to a Windows thread and key presses are passed safely into the FreeRTOS simulator using an integer as a pseudo-interrupt flag that gets polled by a task in the FreeRTOS simulator. This change makes keyboard input stable and safe.
* Instead of triggering behavior from any key being pressed, you now press 't' to output the trace and you press 'r' to reset the timer for the blinky demo.
* Critical sections have been added around printf calls made by FreeRTOS tasks to prevent a potential deadlock happening if the FreeRTOS scheduler would cause a context switch while a task calls printf. This improves the overall stability of the demo immensely.
Test Steps
-----------

Built and run both the full and blinky demo in Visual Studio 2019. Additionally, verified output traces.

Related Issue
-----------

### DIFF
--- a/FreeRTOS/Demo/WIN32-MSVC/FreeRTOSConfig.h
+++ b/FreeRTOS/Demo/WIN32-MSVC/FreeRTOSConfig.h
@@ -46,7 +46,7 @@
 #define configUSE_DAEMON_TASK_STARTUP_HOOK		1
 #define configTICK_RATE_HZ						( 1000 ) /* In this non-real time simulated environment the tick frequency has to be at least a multiple of the Win32 tick frequency, and therefore very slow. */
 #define configMINIMAL_STACK_SIZE				( ( unsigned short ) 70 ) /* In this simulated case, the stack only has to hold one small structure as the real stack is part of the win32 thread. */
-#define configTOTAL_HEAP_SIZE					( ( size_t ) ( 100 * 1024 ) ) /* This demo tests heap_5 so places multiple blocks within this total heap size.  See mainREGION_1_SIZE to mainREGION_3_SIZE definitions in main.c. */
+#define configTOTAL_HEAP_SIZE					( ( size_t ) ( 49 * 1024 ) ) /* This demo tests heap_5 so places multiple blocks within this total heap size.  See mainREGION_1_SIZE to mainREGION_3_SIZE definitions in main.c. */
 #define configMAX_TASK_NAME_LEN					( 12 )
 #define configUSE_TRACE_FACILITY				1
 #define configUSE_16_BIT_TICKS					0
--- a/FreeRTOS/Demo/WIN32-MSVC/FreeRTOSConfig.h
+++ b/FreeRTOS/Demo/WIN32-MSVC/FreeRTOSConfig.h
@@ -46,7 +46,7 @@
 #define configUSE_DAEMON_TASK_STARTUP_HOOK		1
 #define configTICK_RATE_HZ						( 1000 ) /* In this non-real time simulated environment the tick frequency has to be at least a multiple of the Win32 tick frequency, and therefore very slow. */
 #define configMINIMAL_STACK_SIZE				( ( unsigned short ) 70 ) /* In this simulated case, the stack only has to hold one small structure as the real stack is part of the win32 thread. */
-#define configTOTAL_HEAP_SIZE					( ( size_t ) ( 54 * 1024 ) ) /* This demo tests heap_5 so places multiple blocks within this total heap size.  See mainREGION_1_SIZE to mainREGION_3_SIZE definitions in main.c. */
+#define configTOTAL_HEAP_SIZE					( ( size_t ) ( 100 * 1024 ) ) /* This demo tests heap_5 so places multiple blocks within this total heap size.  See mainREGION_1_SIZE to mainREGION_3_SIZE definitions in main.c. */
 #define configMAX_TASK_NAME_LEN					( 12 )
 #define configUSE_TRACE_FACILITY				1
 #define configUSE_16_BIT_TICKS					0
--- a/FreeRTOS/Demo/WIN32-MSVC/FreeRTOSConfig.h
+++ b/FreeRTOS/Demo/WIN32-MSVC/FreeRTOSConfig.h
@@ -46,7 +46,7 @@
 #define configUSE_DAEMON_TASK_STARTUP_HOOK		1
 #define configTICK_RATE_HZ						( 1000 ) /* In this non-real time simulated environment the tick frequency has to be at least a multiple of the Win32 tick frequency, and therefore very slow. */
 #define configMINIMAL_STACK_SIZE				( ( unsigned short ) 70 ) /* In this simulated case, the stack only has to hold one small structure as the real stack is part of the win32 thread. */
-#define configTOTAL_HEAP_SIZE					( ( size_t ) ( 52 * 1024 ) ) /* This demo tests heap_5 so places multiple blocks within this total heap size.  See mainREGION_1_SIZE to mainREGION_3_SIZE definitions in main.c. */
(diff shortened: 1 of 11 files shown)


alfred2g (dismissed):

FreeRTOS/Demo/WIN32-MSVC/Trace_Recorder_Configuration/trcConfig.h
```diff
@@ -1,55 +1,19 @@
...
- * www.percepio.com
- ******************************************************************************/
+ */
 
 #ifndef TRC_CONFIG_H
 #define TRC_CONFIG_H
 
 #ifdef __cplusplus
-extern "C" {
+    extern "C" {
```
paulbartell, line 16: Nitpick: This should be wrapped with uncrustify ignore flags.

FreeRTOS/Demo/WIN32-MSVC/Trace_Recorder_Configuration/trcConfig.h
```diff
@@ -59,12 +23,11 @@ extern "C" {
  * required at least for the ARM Cortex-M port, that uses the ARM CMSIS API.
  * Try that in case of build problems. Otherwise, remove the #error line below.
  *****************************************************************************/
-//#error "Trace Recorder: Please include your processor's header file here and remove this line."
+/*#error "Trace Recorder: Please include your processor's header file here and remove this line." */
```
paulbartell, line 26: Line should be removed?

FreeRTOS/Demo/WIN32-MSVC/main.c
```diff
@@ -138,272 +146,301 @@ static BaseType_t xTraceRunning = pdTRUE;
...
+    volatile uint32_t ulSetToNonZeroInDebuggerToContinue = 0;
+
+    /* Called if an assertion passed to configASSERT() fails.  See
+     * http://www.freertos.org/a00110.html#configASSERT for more information. */
+
+    /* Parameters are not used. */
+    ( void ) ulLine;
+    ( void ) pcFileName;
+
+    printf( "ASSERT! Line %ld, file %s, GetLastError() %ld\r\n", ulLine, pcFileName, GetLastError() );
```
paulbartell, line 287: Printf should probably be called inside the critical section

paulbartell (dismissed):
A couple small things but mostly LGTM.

alfred2g (dismissed):

FreeRTOS/Demo/WIN32-MSVC/main_blinky.c
```diff
@@ -155,7 +161,7 @@ const TickType_t xTimerPeriod = mainTIMER_SEND_FREQUENCY_MS;
 		/* Create the software timer, but don't start it yet. */
 		xTimer = xTimerCreate( "Timer",				/* The text name assigned to the software timer - for debug only as it is not used by the kernel. */
 								xTimerPeriod,		/* The period of the software timer in ticks. */
-								pdFALSE,			/* xAutoReload is set to pdFALSE, so this is a one-shot timer. */
+								pdTRUE,			    /* xAutoReload is set to pdTRUE, so this is timer repeatedly goes off. */
```
paulbartell, line 164: Rephrase?

FreeRTOS/Demo/WIN32-MSVC/main_blinky.c
```diff
@@ -237,37 +243,63 @@ uint32_t ulReceivedValue;
...
+            /* Critical section around printf to prevent a deadlock
+               on context switch. */
+            taskENTER_CRITICAL();
+            {
+                printf("\r\nResetting software timer.\r\n\r\n");
+            }
+            taskEXIT_CRITICAL();
+
+            /* Reset the software timer. */
+            xTimerReset( xTimer, portMAX_DELAY );
```
Skptak, line 295: Is this intentionally left out of the task critical portion?

jasonpcarroll: hmm. it isn't necessary to include it in it. the reason the critical section is there in the first place is just to wrap the printf to prevent a context switch while calling it as that will cause deadlock if another task calls printf before this finishes. it wouldn't hurt it if it was but this is fine.

FreeRTOS/Demo/WIN32-MSVC/FreeRTOSConfig.h
```diff
@@ -46,7 +46,7 @@
 #define configUSE_DAEMON_TASK_STARTUP_HOOK		1
 #define configTICK_RATE_HZ						( 1000 ) /* In this non-real time simulated environment the tick frequency has to be at least a multiple of the Win32 tick frequency, and therefore very slow. */
 #define configMINIMAL_STACK_SIZE				( ( unsigned short ) 70 ) /* In this simulated case, the stack only has to hold one small structure as the real stack is part of the win32 thread. */
-#define configTOTAL_HEAP_SIZE					( ( size_t ) ( 52 * 1024 ) ) /* This demo tests heap_5 so places multiple blocks within this total heap size.  See mainREGION_1_SIZE to mainREGION_3_SIZE definitions in main.c. */
+#define configTOTAL_HEAP_SIZE					( ( size_t ) ( 100 * 1024 ) ) /* This demo tests heap_5 so places multiple blocks within this total heap size.  See mainREGION_1_SIZE to mainREGION_3_SIZE definitions in main.c. */
```
Skptak, line 49: So I set mine to use this "100 * 1024" value as well since I also needed to increase it, was wondering if you had done any testing about this value or if you had just increase it?

jasonpcarroll: My demo was running out of heap. so i just added more.

FreeRTOS/Demo/WIN32-MSVC/Trace_Recorder_Configuration/trcConfig.h
```diff
@@ -198,72 +121,102 @@ extern "C" {
...
- * If this is zero (0), the trace will exclude any Timer events.
+ * This controls how many tasks that can be monitored by the stack monitor.
+ * If this is too small, some tasks will be excluded and a warning is shown.
  *
- * Default value is 0 since dependent on timers.c
- *****************************************************************************/
-#define TRC_CFG_INCLUDE_TIMER_EVENTS 1
+ * Default value is 10.
+ */
+#define TRC_CFG_STACK_MONITOR_MAX_TASKS       50
```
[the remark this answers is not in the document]
jasonpcarroll: honestly - i think this is more for the stream recorder. the FreeRTOS demo makes upwards of 40 tasks so I figured I would set this appropriately.

FreeRTOS/Demo/WIN32-MSVC/main.c
```diff
@@ -53,33 +53,44 @@
...
-#define mainREGION_1_SIZE	8201
-#define mainREGION_2_SIZE	31905
-#define mainREGION_3_SIZE	7807
+ * that make up the total heap.  heap_5 is only used for test and example purposes
+ * as this demo could easily create one large heap region instead of multiple
+ * smaller heap regions - in which case heap_4.c would be the more appropriate
+ * choice.  See http://www.freertos.org/a00111.html for an explanation. */
+#define mainREGION_1_SIZE                     8201
+#define mainREGION_2_SIZE                     60905
+#define mainREGION_3_SIZE                     17807
```
paulbartell, line 85: These three do not add up to 100k specified in configTOTAL_HEAP_SIZE.

Skptak (approved):

paulbartell (approved):

PR #897, opened 2022-12-08
Fix MingW RTOS demo.

jasonpcarroll:
Description
-----------

This PR fixes the WIN32-MingW demo in FreeRTOS/Demo and adds the following changes to functionality:
* Trace recorder continues after saving the Trace output to a file rather than just stopping altogether after a trace dump, allowing users to output / update the trace as many times as they want while the demo runs.
* Keyboard input handling has been moved to a Windows thread and key presses are passed safely into the FreeRTOS simulator using a simulated interrupt.
* Instead of triggering behavior from any key being pressed, you now press 't' to output the trace and you press 'r' to reset the timer for the blinky demo.
* Critical sections have been added around printf calls made by FreeRTOS tasks to prevent a potential deadlock happening if the FreeRTOS scheduler would cause a context switch while a task calls printf. This improves the overall stability of the demo immensely.
Test Steps
-----------

Built and run both the full and blinky demo in using Eclipse IDE 2022-09. Additionally, verified output traces.

Related Issue
-----------


paulbartell:
Could you add a makefile for those of us who prefer not to use eclipse?

FreeRTOS/Demo/WIN32-MingW/main.c
```diff
@@ -134,204 +159,221 @@ declared here, as a global, so it can be checked by a test that is implemented
...
+    is only used for test and example reasons.  Heap_4 is more appropriate.  See
+    http://www.freertos.org/a00111.html for an explanation. */
+    prvInitialiseHeap();
+
+    /* Do not include trace code when performing a code coverage analysis. */
+    #if( projCOVERAGE_TEST != 1 )
+    {
+        /* Initialise the trace recorder.  Use of the trace recorder is optional.
+        See http://www.FreeRTOS.org/trace for more information. */
+        configASSERT( xTraceInitialize() == TRC_SUCCESS );
```
Skptak, line 202: Is there a reason for the swap to calling xTraceInitialize()  here instead of  the vTraceEnable( TRC_START ) call that we've been using?

FreeRTOS/Demo/WIN32-MingW/main_blinky.c
```diff
@@ -182,84 +186,124 @@ TickType_t xNextWakeTime;
...
+    switch ( xKeyPressed )
+    {
+
+    case mainRESET_TIMER_KEY:
+
+        if ( xTimer != NULL )
+        {
+            /* Critical section around printf to prevent a deadlock
+               on context switch. */
+            portENTER_CRITICAL();
```
Skptak, line 292: Was there a reason for using the portENTER/portEXIT macros over the taskENTER/EXIT ones? Asking since they're used interchangeably in this demo but in the other ones it was only the taskENTER/EXIT ones.

FreeRTOS/Demo/WIN32-MingW/main.c
```diff
@@ -134,204 +159,221 @@ declared here, as a global, so it can be checked by a test that is implemented
...
-			printf( "\r\nFailed to create trace dump file\r\n" );
-		}
-	}
-	#endif
+    /* Tracing is not used when code coverage analysis is being performed. */
+    #if( projCOVERAGE_TEST != 1 )
+    {
+        FILE* pxOutputFile;
+
+        vTraceStop();
```
Skptak, line 359: Looking through the ways this function gets called, does this call to TraceStop() need to be here?
It seems like the ways to end up in this function are wrapped with TraceDisable()/TraceEnable() calls already?

FreeRTOS/Demo/WIN32-MingW/main.c
jasonpcarroll: vTraceEnable is deprecated by the lib. If you look at where vTraceEnbale is defined it says its a deprecated call. so I've updated to the new calls.

FreeRTOS/Demo/WIN32-MingW/main_blinky.c
jasonpcarroll: For winsim they technically do the same thing, I used portENTER_CRITICAL here because this is not a task but an interrupt handler.

FreeRTOS/Demo/WIN32-MingW/main.c
jasonpcarroll: ya that shouldn't be there. doesn't hurt anything if the trace is already stopped, but I will remove it.

jasonpcarroll:
> Could you add a makefile for those of us who prefer not to use eclipse? (CI for example)

Hmm sure.

FreeRTOS/Demo/WIN32-MingW/main.c
Skptak: Swapping to the new "xTraceEnable" would make more sense in that case then right?
I'm mainly asking because [xTraceEnable()](https://github.com/FreeRTOS/FreeRTOS/blob/main/FreeRTOS-Plus/Source/FreeRTOS-Plus-Trace/trcSnapshotRecorder.c#L327) does more than just the [initialize call here](https://github.com/FreeRTOS/FreeRTOS/blob/main/FreeRTOS-Plus/Source/FreeRTOS-Plus-Trace/trcSnapshotRecorder.c#L1669)
If the trace.dump file it is making is working then I suppose that's what matters

FreeRTOS/Demo/WIN32-MingW/main_blinky.c
Skptak: Gotcha, in that case do you want to swap over the task calls to these port ones for the sake of consistency?

FreeRTOS/Demo/WIN32-MingW/main.c
jasonpcarroll: I call both. Technically xTraceEnable does both so I should only call it, but figured I'd stick with this as it feels like the more proper way to do it. Calling just xTraceEnable it could fail for a couple of reasons. If I call xTraceInitialize then xTraceEnable, I know that I failed to initialize rather than failed to start.

FreeRTOS/Demo/WIN32-MingW/main_blinky.c
jasonpcarroll: I use the task calls in tasks as you would normally.

Skptak (approved):

archigup:
LGTM, but Paul's comment does not seem to be addressed

jasonpcarroll:
> LGTM, but Paul's comment does not seem to be addressed

I've added the Makefile for the demo.

archigup (approved):

Skptak (approved):

PR #908, opened 2023-01-05
Add VeriFast thread safety proof for context switches in SMP port

tobireinhard:
Add VeriFast thread safety proof for context switches in SMP port

Description
-----------
The SMP implementation of the scheduler suffers from a buffer underflow, cf. the fix proposed in [PR 607](https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/607). We used [VeriFast](https://github.com/verifast/verifast), a deductive program verifier for C, to prove that once the fix from PR 607 has been applied, context switches are memory safe (i.e. no memory error) and mutually thread safe (i.e. no race condition). 

The proof is an *unbounded* proof. That is, it considers any possible number of tasks and any possible sizes for the involved data structures. Further, it considers any possible task interleavings and any possible interrupt schedule that may occur during runtime. Taking all this into account, it ensure that no memory error and no race condition can occur during runtime, when we run multiple concurrent context switches on different cores.

See the proof README for a more detailed explanation.

Test Steps
-----------
Run one of the scripts `run-verifast.sh` or `run-vfide.sh` as described in the proof README.

Related Issue
-----------
https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/607


gmtt:
Hi @tobireinhard — first, thank you very much for this contribution and for the
substantial effort that went into the VeriFast proof. The work you and the
VeriFast team did on formally verifying the SMP context-switch path was
genuinely valuable, and we appreciate it.

Unfortunately we're going to close this PR without merging. The reasons:

1. This PR was structured around the buffer-underflow fix proposed in
   FreeRTOS-Kernel#607. That kernel PR was closed in January 2023 in favor of
   a different fix in FreeRTOS-Kernel#610, so the proof here no longer
   matches what shipped on `main`.
2. The PR has been a draft and unmerged for ~3 years, is significantly behind
   `main`, and re-validating / re-aligning the proof against the current
   kernel would be a substantial undertaking — comparable to writing a new
   proof — rather than a refresh of this one.
3. Subsequent kernel work has changed the SMP scheduler internals that this
   proof reasons over, so reviving the proof in-place wouldn't give us
   meaningful guarantees about the current code.

If you (or the VeriFast team) are interested in reviving formal verification
of the SMP context-switch path against current `main`, we'd be happy to
discuss it as a fresh contribution — please feel free to open a new issue or
PR and tag the FreeRTOS-Kernel maintainers.

Thanks again for the contribution.

PR #939, opened 2023-03-01
Update Readme for the FreeRTOS_PLUS_TCP_ECHO_QEMU_msp2

kar-rahul-aws:
Update the documentation for Readme of the FreeRTOS_PLUS_TCP_ECHO_QEMU_msp2

Description
-----------
The documentation for the ReadMe for the QEMU demo is updated , with focus on giving a clear understanding to the user of :

1. The various address need to be set as part of configuration
2. The systems needed to run the Demo on (https://forums.freertos.org/t/example-freertos-tcp-echo-qemu-is-not-providing-the-desirable-result/)
3. A pictorial representation on how the QEMU setup functions after the Virtual Bridge is setup.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/FreeRTOS-Plus/Demo/FreeRTOS_Plus_TCP_Echo_Qemu_mps2/Readme.md
+++ b/FreeRTOS-Plus/Demo/FreeRTOS_Plus_TCP_Echo_Qemu_mps2/Readme.md
@@ -1,197 +1,355 @@
-# Emulating MPS2 Cortex M3 AN385 on QEMU
+# TCP Echo Client Demo for MPS2 Cortex-M3 AN385 emulated using QEMU
+This FreeRTOS+TCP example demonstrates a TCP Echo Client which sends
+echo requests to an Echo Server and then receives the echo reply. The
+Echo Client runs on the MPS2 Cortex-M3 AN385 platform emulated using QEMU.
 
 ## Requirements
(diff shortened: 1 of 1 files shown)


FreeRTOS-Plus/Demo/FreeRTOS_Plus_TCP_Echo_Qemu_mps2/Readme.md
```diff
@@ -23,6 +23,46 @@ enp0s3:     ethernet interface
...
 ```
+## Setting up the QEMU Demo
+
+To setup the system for running the QEMU demo, please check these steps.
+```
+1. Setting up the client-server system
+    a. Option-1 : The demo can be run on 2 different systems, one with the client running FreeRTOS system in qemu and another running the echo server.
+    b. Option-2 : The demo could also be run on 1 system, with the QEMU client running on Ubuntu Virtual Machine (VM) and the host OS (Windows/MAC/Linux) running the server.
+
+2. Setting up MAC and IP Address
```
alfred2g, line 34: Could we specify which definitions would hold each? we always get questions from confused customers that map concepts to actual macros they need to fill up

alfred2g (approved):

AniruddhaKanhere (approved):

ActoryOu (changes requested):

aggarg (approved):

ActoryOu (approved):

PR #950, opened 2023-03-07
Update Backward Compatibility Flag

moninom1:
Description
-----------

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/FreeRTOS-Plus/Demo/Common/FreeRTOS_Plus_CLI_Demos/UDP-Related-CLI-commands.c
+++ b/FreeRTOS-Plus/Demo/Common/FreeRTOS_Plus_CLI_Demos/UDP-Related-CLI-commands.c
@@ -266,7 +266,7 @@ uint32_t ulAddress;
 	switch( xIndex )
 	{
 		case 0 :
-		#if defined( FREERTOS_PLUS_TCP_VERSION ) && ( FREERTOS_PLUS_TCP_VERSION >= 10 )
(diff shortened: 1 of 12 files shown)


FreeRTOS-Plus/Demo/Common/FreeRTOS_Plus_CLI_Demos/UDP-Related-CLI-commands.c
```diff
@@ -266,7 +266,7 @@ uint32_t ulAddress;
 	switch( xIndex )
 	{
 		case 0 :
-		#if defined( FREERTOS_PLUS_TCP_VERSION ) && ( FREERTOS_PLUS_TCP_VERSION >= 10 )
+		#if defined( ipconfigIPv4_BACKWARD_COMPATIBLE ) && ( ipconfigIPv4_BACKWARD_COMPATIBLE == 0 )
```
sirnish, line 269: Just curious, If we are introducing IPv4 compatibility flag, then may be we would have similar flag for IPv6 as well for future changes. But would it not introduce more challenges in terms of maintainability.

moninom1: Hi sirnish, for now we are thinking of only 1 flag and after a lot of discussion we decided to make it backward compatible but in most of the cases we are internally calling new APIs from the older APIs to reduce the overhead or maintainability.

PR #997, opened 2023-04-21
Added Posix Demo run in github action

xuelix:
Description
-----------
During the action, after demo build, Posix_GCC port will run both Full demo and Blinky demo.


n9wxu (approved):

n9wxu (dismissed):

.github/workflows/freertos_demos.yml
```diff
@@ -66,10 +68,42 @@ jobs:
...
         working-directory: FreeRTOS/Demo/Posix_GCC
-        run: make -j
+        run: |
+          make -j
+
+      - name: Run and monitor Posic_GCC Full Demo
+        if: success() || failure() && steps.build-posix-gcc-demo.outcome == 'success'
+        working-directory: FreeRTOS/Demo/Posix_GCC
+        run: |
+          timeout 3m build/posix_demo | tee output.txt
```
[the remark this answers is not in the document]
jasonpcarroll: One thing to be aware of with this to is that any command run in the action's- "run" field will fail the action if the exit status is not 0 ... which is a bit of an annoying caveat ...

jasonpcarroll (approved):

.github/workflows/freertos_demos.yml
```diff
@@ -66,10 +68,44 @@ jobs:
           sudo apt-get -y update
           sudo apt-get -y install build-essential
 
-      - name: Build Posix_GCC Demo
+      - name: Build Posix_GCC Full Demo
+        id: build-posix-gcc-demo
```
jasonpcarroll, line 72: Rename this id to build-posix-gcc-full-demo

jasonpcarroll:
```suggestion
        id: build-posix-gcc-full-demo
```

.github/workflows/freertos_demos.yml
```diff
@@ -66,10 +68,44 @@ jobs:
...
         shell: bash
         working-directory: FreeRTOS/Demo/Posix_GCC
-        run: make -j
+        run: |
+          make -j
+
+      - name: Run and monitor Posix_GCC Full Demo
+        if: success()
```
jasonpcarroll, line 79:
```suggestion
        if: success() || failure() && steps.build-posix-gcc-full-demo.outcome == 'success'
```

.github/workflows/freertos_demos.yml
```diff
@@ -66,10 +68,44 @@ jobs:
...
+          timeout 5m build/posix_demo | tee output.txt
+          if grep -q "OK: No errors" output.txt; then
+            echo -e "\nSUCCESS - Found SUCCESS string in output."
+          else
+            echo -e "\nFAILURE - SUCCESS string not found in output."
+            exit 1
+          fi
+
+      - name: Build Posix_GCC Blinky Demo
+        if: success() || failure() && steps.build-posix-gcc-demo.outcome == 'success'
```
jasonpcarroll, line 91:
```suggestion
        id: build-posix-gcc-blinky-demo
        if: success() || failure()
```

.github/workflows/freertos_demos.yml
```diff
@@ -66,10 +68,44 @@ jobs:
...
+        shell: bash
+        working-directory: FreeRTOS/Demo/Posix_GCC
+        run: |
+          sed -i "s/#define[[:space:]]\+mainSELECTED_APPLICATION[[:space:]]\+FULL_DEMO/#define    mainSELECTED_APPLICATION      BLINKY_DEMO/g" main.c
+          make -j
+
+      - name: Run and monitor Posix_GCC Blinky Demo
+        if: success() || failure() && steps.build-posix-gcc-demo.outcome == 'success'
```
jasonpcarroll, line 99:
```suggestion
        if: success() || failure() && steps.build-posix-gcc-blinky-demo.outcome == 'success'
```

.github/workflows/freertos_demos.yml
jasonpcarroll: adding this id will allow you to differentiate between the two make steps. success() || failure() will make this run despite previous steps failing, but will still allow you to cancel the flow. You could add an ID to the install gcc step and add it as a dependency as well but probably not needed

.github/workflows/freertos_demos.yml
jasonpcarroll: This will make it so this only runs as long as the build for the blinky demo succeeds if you commit my previous suggestion.

.github/workflows/freertos_demos.yml
jasonpcarroll: Rename this to this so you can differentiate between the two build steps

jasonpcarroll (changes requested):

jasonpcarroll (approved):

n9wxu (approved):

PR #1018, opened 2023-05-26
Create GitHub Workflows that execute the p3 FreeRTOS Kernel Demos

ChaiTowKway:
Current git workflow only builds demo, but no action to run the demo for testing.
This PR is to update workflow file to run Posix_GCC, WIN32-MSVC and WIN32-MingW demos in git action.
For each build action, the function setvbuf is added into the main.c file to force stdout to write immediately by setting the buffer size for it to 0. This is to allow real time response when detecting success line during run action.

### DIFF
--- a/FreeRTOS/Demo/Posix_GCC/main.c
+++ b/FreeRTOS/Demo/Posix_GCC/main.c
@@ -181,6 +181,7 @@ int main( void )
     #elif ( mainSELECTED_APPLICATION == FULL_DEMO )
         {
             console_print( "Starting full demo\n" );
+            console_print( "Testing full demo\n" );
             main_full();
         }
     #else
--- a/FreeRTOS/Demo/Posix_GCC/main.c
+++ b/FreeRTOS/Demo/Posix_GCC/main.c
@@ -152,6 +152,9 @@ static clockid_t cid = CLOCK_THREAD_CPUTIME_ID;
 
 int main( void )
 {
+    /* No buffer for stdout. */
+    setvbuf(stdout, NULL, _IONBF, 0);
+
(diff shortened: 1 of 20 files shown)


FreeRTOS/Demo/Posix_GCC/main.c
```diff
@@ -152,6 +152,9 @@ static clockid_t cid = CLOCK_THREAD_CPUTIME_ID;
 
 int main( void )
 {
+    /* No buffer for stdout. */
+    setvbuf(stdout, NULL, _IONBF, 0);
```
Skptak, line 156: Formatting issue

FreeRTOS/Demo/WIN32-MSVC-Static-Allocation-Only/main.c
```diff
@@ -86,6 +86,9 @@ static void prvCheckTask( void *pvParameters );
 
 int main( void )
 {
+	/* No buffer for stdout. */
+    setvbuf(stdout, NULL, _IONBF, 0);
```
Skptak, line 90: Formatting issue, need spaces after a '(' character and before a ')' character

ChaiTowKway: Add the function in git workflow file instead of demo main.c file

FreeRTOS/Demo/Posix_GCC/main.c
ChaiTowKway: Formatting error corrected; And the function is added in git workflow file instead of demo main.c file

Skptak (approved):

n9wxu (approved):

PR #1036, opened 2023-07-14
Uncrustify PR Check Changes

Skptak:
Uncrustify PR Check Changes

Description
-----------
Put the uncrustify parsing in a log group
Print out a git diff of the changes if a file is incorrectly formatted so end user can see it

Test Steps
-----------
Run the CI-Checker GitHub Action
[Working Run One](https://github.com/Skptak/FreeRTOS/actions/runs/5549759684/jobs/10134187166#step:4:1), [Working Run Two](https://github.com/Skptak/FreeRTOS/actions/runs/5549811029/jobs/10134289515#step:4:1), Forced a failure on [this run](https://github.com/Skptak/FreeRTOS/actions/runs/5549825239/jobs/10134319642#step:4:290) to show how it looks now

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,29 +33,29 @@ jobs:
         run: sudo apt-get install uncrustify=0.69.0+dfsg1-1build1
       - name: Run Uncrustify
         run: |
-          : # Uncrustify on C files while ignoring symlinks.
-          : # Make a collapsible section in the log to run uncrustify
+          # Uncrustify on C files while ignoring symlinks.
+          # Make a collapsible section in the log to run uncrustify
           echo "::group::Uncrustify Check"
-          : # GitHub automtically use "set -e" which causes scripts to fail on the first exit code
-          : # This would mean the first time a file fails the check that we exit without formatting all files.
+          # GitHub automtically use "set -e" which causes scripts to fail on the first exit code
+          # This would mean the first time a file fails the check that we exit without formatting all files.
           set +e
           uncrustify --version
           find FreeRTOS/Demo/Common FreeRTOS/Test \( -name ethernet -o -name drivers -o -path 'FreeRTOS/Test/CMock/CMock' \) -prune -false -o -name "*.[hc]" | xargs uncrustify --no-backup --replace --if-changed -c tools/uncrustify.cfg -l C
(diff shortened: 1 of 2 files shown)


jasonpcarroll:
I like this change - should we add it to the github action in CI-CD-GitHub-Actions then swap this out for the action? I've been trying to push for uncrustify to have a single source of truth.

A-Zaba (approved):

.github/workflows/ci.yml
```diff
@@ -33,26 +33,46 @@ jobs:
...
           fi
 
           find FreeRTOS/Demo/Common   \( -name "ethernet" \) -prune -o -false -o -type f -a -name "*" -exec grep --color=yes -In -e "[[:blank:]]$" {} +
           if [ "$?" = "0" ]; then
-            echo "Files have trailing whitespace."
+            echo -e "\033[32;31mFiles have trailing whitespace.\033[0m"
             exit 1
           else
             if [ "$ERROR" -eq "1" ]; then
+              echo -e "\033[32;31mFiles have trailing whitespace.\033[0m"
```
kstribrnAmzn, line 72: I'm a little confused on why this echo was introduced? Wouldn't the echo statement on line 68 handle this case?

kstribrnAmzn (approved):

.github/workflows/ci.yml
Skptak: This handles the case from like 61/62. It does two seperate checks for trailing whitespace. So line 68 prints if the second check hits a trailing whitespace character while this line prints if the first check hits a trailing whitespace character,

AniruddhaKanhere: Can we merge these two separate lines into one check like so:
```
if [ "$?" = "0" ] || [ "$ERROR" -eq "1" ]; then
```

That should work I believe.

AniruddhaKanhere (approved):
Approved with a minor comment.

.github/workflows/ci.yml
Skptak: Yeah that would absolutely work. Since this has the two approvals already I'm going to merge it as it is. We know we want to update all these formatting actions in the near future where stuff like this could be fixed then.

PR #1054, opened 2023-07-27
Create GitHub workflows to execute the WinSim demos with libslirp network interface implemented in FreeRTOS+TCP VS Static Project

ChaiTowKway:
Description
-----------
1. Add Libslirp in FreeRTOS+TCP VS Static Project for existing FreeRTOS+TCP demos in P3 to swap WinPCap with libslirp for user mode networking
2. Create GitHub workflows that execute the FreeRTOS+ demos using these configurations with libslirp on GitHub Runners
3. Connect to Echo server, local MQTT and HTTP server, AWS IOT & S3 service in GitHub workflows for corresponding FreeRTOS+TCP demo execution

Test Steps
-----------

Checklist:
----------

- [X] I have tested my changes. No regression in existing tests.

Related Issue
-----------


FreeRTOS-Plus/VisualStudio_StaticProjects/FreeRTOS+TCP/resource.h
Skptak, line 1: Is this file used for anything?

ChaiTowKway: Not being used. Removed the file from the static project folder.

Skptak (approved):

paulbartell (approved):
Would prefer that we build a minimal set of glib and separate libslirp and glib into their own vcxproj files, but this should be good enough.

PR #1112, opened 2023-11-08
Fix one typical misusage of CMock.

Moral-Hao:
Description
-----------

CMock has three policies when dealing with pointer:
1. compare_ptr : compare the pointer itself;
2. compare_data : compare the data which is pointed by pointer;
3. smart : compare the pointer itself, or compare the data; 

By default, CMock use compare_data policy.

Most test cases in FreeRTOS need to compare the pointer itself, and we haven't use CMock properly.
For example, one test case of vTaskResume is wrong, and because CMock compares the data pointed by pointer currently, so the error didn't exposed.

There are so many misusage need to fix, it will take a lot of work. I'm here to fix one typical misusage first.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/FreeRTOS/Test/CMock/tasks/tasks.yml
+++ b/FreeRTOS/Test/CMock/tasks/tasks.yml
@@ -1,6 +1,7 @@
 :cmock:
   :mock_prefix: mock_
   :when_no_prototypes: :warn
+  :when_ptr: :smart
   :treat_externs: :include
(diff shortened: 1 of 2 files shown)


cookpate (approved):
Thank you for the PR. LGTM

n9wxu (approved):

PR #1120, opened 2023-11-22
Remove 64 Bit Visual Studio Configurations

Skptak:
Description
-----------
The Visual Studio Projects have a configuration option for "Release" and for "x64". These are not actively worked on or maintained. These configurations were generated when creating the project structure with the static library visual projects. Additionally, they all have the `WIN32` flag set in the pre-processor, leading to potential issues when run as a 64 bit application. As such they should be removed from the repository, so as not to mislead people who may attempt to run them. 

Test Steps
-----------
Removed the configuration options. Then built the demos locally to ensure the projects built.
Used GitHub CI-CD as well as internal CI-CD to ensure the demos run correctly.

Checklist:
----------

Related Issue
-----------


paulbartell (approved):

FreeRTOS/Demo/WIN32-MingW/Trace_Recorder_Configuration/trcConfig.h
```diff
@@ -1,319 +1,333 @@
...
+
+/** Exclude Windows APIs such as Cryptography, DDE, RPC, Shell,
+* and Windows Sockets. */
+#ifndef WIN32_LEAN_AND_MEAN
+    #define WIN32_LEAN_AND_MEAN
+#endif
+
+#include <winsock2.h>
+
+/* Define _WINSOCKAPI_ to ensure that winsock.h is not included */
```
paulbartell, line 27: Does this need to be defined before the include?

FreeRTOS-Plus/Demo/corePKCS11_MQTT_Mutual_Auth_Windows_Simulator/corePKCS11_MQTT_Mutual_Auth.vcxproj.filters
```diff
@@ -72,6 +72,7 @@
     <ClCompile Include="PKCS11_Mqtt_MutualAuthDemo.c">
       <Filter>Source</Filter>
     </ClCompile>
+    <ClCompile Include="..\..\Source\Application-Protocols\network_transport\mbedtls_rng_pkcs11.c" />
```
bradleysmith23, line 75: Is there a reason this does not have a filter associated with it?

Skptak: I honestly have no idea, this is a file generated by Visual Studio, I didn't personally do anything different here

bradleysmith23 (approved):
Think this change makes a lot of sense (especially as someone who spent a lot of time debugging the demos recently, only to realize that the issue was using the x64 config not Win32). It is pretty painful the way it is currently, especially since it is not easy to identify this problem.

PR #1121, opened 2023-11-24
Fix misusage of CMock in dir tasks.

Moral-Hao:
Change CMock pointer policy of tasks from `smart` to `compare_ptr`.

Description
-----------

Just like the previous [PR#1112](https://github.com/FreeRTOS/FreeRTOS/pull/1112) said,  we need to compare the pointer itself instead of compare the data pointed by the pointer.
CMock has two policies which can achieve it:

1. `smart` : with this policy, we need to change all the `xxx_Expect` to `xxx_ExpectWithArray`, and add an extra argument  for every pointer;
2. `compare_ptr` : with this policy, we don't need to change `xxx_Expect`, but we need to remove the CMock plugins `array`(because they are conflict options, and we haven't use feature `array`).

To minimize the change, I chose policy 2 `compare_ptr` here.
And I fixed the errors of test cases found by this change.

Test Steps
-----------

`cd ./FreeRTOS/Test/CMock`
`make -C tasks/ lcov`

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/FreeRTOS/Test/CMock/tasks/tasks.yml
+++ b/FreeRTOS/Test/CMock/tasks/tasks.yml
@@ -1,7 +1,7 @@
 :cmock:
   :mock_prefix: mock_
   :when_no_prototypes: :warn
-  :when_ptr: :smart
+  :when_ptr: :compare_ptr
   :treat_externs: :include
   :treat_inlines: :include
   :enforce_strict_ordering: TRUE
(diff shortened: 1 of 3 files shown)


xuelix:
Thanks for the PR.
We'll have a pr-br reviewed the changes.

paulbartell (approved):

n9wxu (approved):

PR #1136, opened 2023-12-13
Fix formatting in kernel demo application files and header files

kar-rahul-aws:
Description
-----------
This PR cleans up the kernel demo application files, using uncrustify tool, so that all the application and header files for FreeRTOS+Kernel Demos use spaces.

Test Steps
-----------
NA

Checklist:
----------


FreeRTOS/Demo/AVR_ATMega323_WinAVR/regtest.c
```diff
@@ -45,312 +45,311 @@ portBASE_TYPE xRegTestError = pdFALSE;
...
+        asm ( "LDI	r19,	20");
+        asm ( "LDI	r20,	21");
+        asm ( "LDI	r21,	22");
+        asm ( "LDI	r22,	23");
+        asm ( "LDI	r23,	24");
+        asm ( "LDI	r24,	25");
+        asm ( "LDI	r25,	26");
+        asm ( "LDI	r26,	27");
+        asm ( "LDI	r27,	28");
+        asm ( "LDI	r30,	29");
```
Skptak, line 265: Is this space between the asm and the paren character still valid?
Formatting these demo files with uncrustify is a bit tricky, as seen in the [FreeRTOS-Kernel PR #768](https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/768) to format the portable directory.


I think it might make sense to start by just trimming whitespace and swapping line endings to use `LF` instead of `CLRF`, which I started in #1081 and then follow that up with formatting files that don't contain the `asm(` string?

kar-rahul-aws: I see the point, and I agree this would cause a problem for the files containing assembly code.
Let me start by raising changes only for the "main" application files and then check the rest.

kar-rahul-aws:
Will raise the PR later with only the main files of the kernel demos

PR #1154, opened 2024-01-11
Update GitPython version for untrusted search path on Windows.

ActoryOu:
Update GitPython version for untrusted search path on Windows.

Description
-----------

From [security dependabot alerts #16](https://github.com/FreeRTOS/FreeRTOS/security/dependabot/16), we should update GitPython version to 3.1.41 to fix untrusted search path under some conditions on Windows.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/.github/scripts/common/requirements.txt
+++ b/.github/scripts/common/requirements.txt
@@ -1,5 +1,5 @@
 Deprecated
-GitPython
+GitPython>=3.1.41
 PyGithub
 PyJWT
 PyYAML
--- a/.github/scripts/release-requirements.txt
+++ b/.github/scripts/release-requirements.txt
@@ -1,5 +1,5 @@
 Deprecated
-GitPython
+GitPython>=3.1.41
 PyGithub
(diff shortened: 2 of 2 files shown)


Skptak:
From what I can tell, Depend-A-Bot alerts don't inspect the versions of libraries used. A longer discussion about this was had in this [PR](https://github.com/FreeRTOS/CI-CD-Github-Actions/pull/89)

The GitHub action already pulls GitPython 3.4.1 as of now without this change. A link to the "Check File Headers" action doing this can be found [here](https://github.com/FreeRTOS/FreeRTOS/actions/runs/7466790818/job/20386184157#step:6:23)

I'm approving this PR as it is using the `GitPython>=3.1.41` instead of `GitPython==3.1.41`. This way we will pull newer versions of the libraries, but will always use the specified minimum version.

Skptak (approved):

amazonKamath (approved):

PR #1237, opened 2024-07-17
fix AbortDelay.c Test

NikJen26:
Fixing Issue with AbortDelay Tests in Release build.

Description
-----------
When xStreamBufferStruct gets out of scope it gets delete and results in an error down the line. This is fixed by moving the variable outside the inner scope. The variable does not need to be static because the Buffer is only used once and is deleted afterwards (per functioncall)

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/FreeRTOS/Demo/Common/Minimal/AbortDelay.c
+++ b/FreeRTOS/Demo/Common/Minimal/AbortDelay.c
@@ -488,7 +488,9 @@
         uint8_t uxRxData;
 
         #if ( configSUPPORT_STATIC_ALLOCATION == 1 )
-            /* The variable used to hold the stream buffer structure. */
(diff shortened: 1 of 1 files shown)


moninom1:
Hi @NikJen26 Thank you for your contribution. Can you please help me understand which test cases are failing without this change?

NikJen26:
In the function `prvTestAbortingStreamBufferReceive` the call to xStreamBufferReceive results in a assert. This is only the case when i enable -02 and -flto.

moninom1:
@NikJen26 Thank you for the info, i would also like to see what error you are getting. Can you please help in sharing the error information.

aggarg (dismissed):

aggarg (approved):

PR #1332, opened 2025-03-04
Demo/CORTEX_MPS2_QEMU_IAR_GCC: increase min stack size from 80 to 88

laroche:
Description
-----------

Compiling this demo with picolibc 1.8.9 (as packaged with Debian trixie) results in a stack overflow. Increasing the minimal stack size from 80 to 88 bytes resolves this. (Debian trixie arm crosscompiler defaults to compile/link against picolibc and not newlib anymore.)

Test Steps
-----------
stack overflow happens directly on test start.

Checklist:
----------

Related Issue
-----------


aggarg (approved):

bhoomrs (approved):
Adapts the demo for modern Debian ARM toolchains using picolibc instead of newlib; increase is only 8 bytes, making it a low-risk change, hence, approving the change.

PR #1381, opened 2025-12-22
docs: add beginner-friendly FreeRTOS overview

Chintan0801:
Added a short beginner-focused overview explaining basic FreeRTOS concepts
such as tasks, scheduler, and tick, to improve accessibility for new users.


xuelix:
The one-liners you added does not seem like a good fit in this README.md.  In the Getting Started section, it has links to more comprehensive resources for beginners.

Chintan0801:
Thanks for the clarification — that makes sense.

I agree that the README is intended to stay high-level and point users to comprehensive beginner resources rather than duplicating content. I’m happy to remove this section or adjust it to better align with the existing documentation structure.

Please let me know which you’d prefer.

PR #1406, opened 2026-04-24
Fix undefined behavior in mbedTLS transport log format specifiers

aniruddhajoshi876:
Description
-----------
Fix three undefined behavior bugs in the mbedTLS transport logging code.

1. `transport_mbedtls.c` and `transport_mbedtls_pkcs11.c`: `psa_crypto_init()`
   returns `psa_status_t`, which is an integer error code. Both files logged
   this value with `%s`, causing the runtime to treat the integer bit pattern
   as a pointer and attempt to read a string from that address. Changed to `%d`
   with an explicit `( int )` cast.

2. `transport_mbedtls.c`: The TLS close-notify log call had its two variadic
   arguments swapped relative to the format string. `%p` was receiving a string
   literal and `%s` was receiving `pNetworkContext`, causing the network context
   struct to be dereferenced as a C string. Corrected argument order and added
   a `( void * )` cast on the pointer argument for strict standard conformance.

Test Steps
-----------
No runtime test path is needed cuz the changes are in logging calls and are
verified by inspection against the C standard for variadic format specifiers.
Build with `-Wformat` to confirm no format warnings on the affected lines.

Checklist:
----------

Related Issue
-----------
None.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,8 +1,0 @@
-{
-  "permissions": {
-    "allow": [
-      "Bash(git submodule *)",
-      "Bash(git checkout *)"
-    ]
-  }
-}
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,8 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(git submodule *)",
+      "Bash(git checkout *)"
+    ]
+  }
+}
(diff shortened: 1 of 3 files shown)


AniruddhaKanhere (approved):
Thank you for your contribution @aniruddhajoshi876!
These changes look good to me. Approved.

aggarg (approved):

aniruddhajoshi876:
> Thank you for your contribution @aniruddhajoshi876! These changes look good to me. Approved.

I checked the failing Windows Simulator CI job. It seems to fail during the “Install Glib” setup step with:
Current directory is not a meson build directory
This looks like a CI/glib/meson setup issue rather than a failure in my changes. Should I rerun the check, or is this a known flaky CI issue?

AniruddhaKanhere:
Thank you @aniruddhajoshi876!

I am merging this in - we will fix the issue in CI/CD. You are welcome to raise a PR if you want :)

Thanks

PR #1407, opened 2026-04-25
fix: replace bare except with except Exception in mpfs_configuration_generator.py

srpatcha:
Automated contribution by @srpatcha.

## Changes
fix: replace bare except with except Exception in mpfs_configuration_generator.py

Signed-off with GPG.


archigup:
We have a backoff library already.
The commit that says prevent double free adds a free?

srpatcha:
Closing per @archigup's feedback — the backoff library already exists and the other changes were unrelated. Reopening as a focused 1-line fix for the bare except clause.

PR #1425, opened 2026-07-24
Demo/BlockQ: add NULL checks after pvPortMalloc

ARYA-mgc:
## Description

In `vStartBlockingQueueTasks()` (`FreeRTOS/Demo/Common/Minimal/BlockQ.c`), all six `pvPortMalloc` calls are dereferenced right after allocation without checking for `NULL`. On memory-constrained targets, if the heap is exhausted, this could lead to a NULL pointer dereference and a hard fault.

## Changes

- Added `configASSERT( ptr != NULL )` after each `pvPortMalloc` call, so failures are caught early in debug builds.
- Wrapped the setup and task-creation code for each allocation in `if( ptr != NULL )` guards, so if an allocation fails, the corresponding tasks are simply not created instead of crashing.
- The second allocation in each pair (e.g., `pxQueueParameters2`) is nested inside the first one's guard, so it's never touched unless the first allocation succeeded.

## Rationale
`semtest.c` already checks `pvPortMalloc` results before use (line 96), so this brings `BlockQ.c` in line with that existing pattern for consistency.

## Testing

- Compiles cleanly.
- No change to the happy path  behavior is identical when all allocations succeed
- On allocation failure, the affected tasks are skipped instead of causing undefined behavior.


ARYA-mgc:
@rawalexe Hi! Could you please review this PR whenever you get a chance .Sorry for tagging you directly I wasn't sure who to reach

PR #1426, opened 2026-07-27
Adapt SMP target tests to configurable test framework

chinglee-iot:
Introduce a configurable test framework for FreeRTOS/Test/Target and adapt existing tests to it

Description
-----------

Adds `tests/include/test_default_setting_config.h`, which defines a small set
of `test*` macros (`testSETUP_FUNCTION_PROTOTYPE`, `testTEARDOWN_FUNCTION_PROTOTYPE`,
`testENTRY_FUNCTION_PROTOTYPE`, `testBEGIN_FUNCTION`, `testRUN_TEST_CASE_FUNCTION`,
`testEND_FUNCTION`) that default to plain Unity calls. Test files use these
macros instead of calling Unity APIs directly, so a target can swap in a
different test framework by providing its own `test_setting_config.h` and
setting `configTARGET_TEST_USE_CUSTOM_SETTING` to `1` in its `FreeRTOSConfig.h`,
without touching any test source file.

Changes:
- Update all existing SMP tests under `tests/smp/` and the `tests/smp/template`
  to use the new `test*` macros instead of calling Unity (`UNITY_BEGIN`,
  `RUN_TEST`, `UNITY_END`, raw `setUp`/`tearDown`/`vRunXxx` prototypes)
  directly.
- Move `README.md` from `tests/smp/template/` to the `Test/Target` root and
  document the test framework (macro table, override mechanism) alongside
  the existing "add a new test/target" instructions, since it now applies
  repo-wide rather than just to the template.

Test Steps
-----------

- Build and run the existing SMP target tests unmodified (default Unity path)
  and confirm they still pass with no behavior change.

Checklist:
----------

Related Issue
-----------


aggarg:
@chinglee-iot Please take a look at the following patch: [suggestions.patch](https://github.com/user-attachments/files/31879820/suggestions.patch).

A target that wants to override test macro definitions can do the following:
```c
#define configTARGET_TEST_USE_CUSTOM_MACRO_DEFINITIONS 1

#define testRUN_TEST_CASE_FUNCTION  CUSTOM_RUN_TEST
#define testBEGIN_FUNCTION          CUSTOM_BEGIN
#define testEND_FUNCTION            CUSTOM_END
```

What do you think about it?

PR #1434, opened 2026-08-31
Make the Common/Minimal tests SMP safe, and add an Espressif ESP32 demo

sudeep-mohanty:
## Description

Makes the `Common/Minimal` tests safe to run on an SMP port, and adds an ESP-IDF demo that runs the whole bar on Espressif targets.

The tests were written for a uniprocessor, so several of them observe a peer task's state immediately after acting on it, or assume a strict priority serialisation. Both assumptions break once tasks run concurrently on more than one core (`configRUN_MULTIPLE_PRIORITIES=1`). Each fix is scoped to the case that is actually unsafe and is guarded so single-core behaviour is unchanged.

Commits are one per test, followed by the demo:

- 9 commits, one per `Common/Minimal` file
- 1 commit adding `Demo/Espressif_ESP32`

## What the two kinds of change are

**SMP port defects**:

- `IntQueue.c` creates its queues before the tasks that use them. The tasks start by accessing those queues, and the first task created runs on the other core while `vStartInterruptQueueTasks()` is still executing.
- `AbortDelay.c` creates its two tasks with the scheduler suspended. Each task starts by looking the other up by name, which cannot work until both exist.
- `StaticAllocation.c` gives the timer ID variable and the task control block static storage. `xTimerDelete()` only queues the delete, and `vTaskDelete()` does not tear down a task that is running on another core, so both were left referenced after their stack frame had been reused.
- `QueueSet.c` and `StreamBufferDemo.c` serialise state that the tick hook and a task both touch and that the object's own locking does not cover.

**Assertions that cannot hold once tasks run concurrently**:

- Peer-state reads via `eTaskGetState()` in `recmutex.c`, `IntSemTest.c`, `GenQTest.c` and `IntQueue.c` are restricted to `configNUMBER_OF_CORES == 1`. `configRUN_MULTIPLE_PRIORITIES == 0` does **not** restore these: priority inheritance raises the mutex holder to the waiter's priority inside `xSemaphoreTake()` before the waiter blocks, and in `IntQueue.c` the low priority task raises its own priority, so in both cases the two tasks are at the same priority and legitimately run at the same time.
- `StaticAllocation.c`'s expiry-count and created-task-state checks use `( configNUMBER_OF_CORES == 1 ) || ( configRUN_MULTIPLE_PRIORITIES == 0 )`, because those premises really do rest on tasks of *different* priorities.

The gate therefore differs per check, by which premise it depends on.

## Relationship to the kernel PR

This is a complementary PR to FreeRTOS/FreeRTOS-Kernel#1154, which adds granular (per-data-group) locking to the SMP kernel.

## Test Steps

    cd FreeRTOS/Demo/Espressif_ESP32
    idf.py -DSDKCONFIG_DEFAULTS="sdkconfig.defaults;sdkconfig.defaults.esp32;sdkconfig.ci.smp_multi_prio" build flash monitor

Each cycle prints a `test_<Name>:PASS|FAIL` line per test. The demo requires `CONFIG_FREERTOS_SMP=y`: `Common/Minimal` calls `portENTER_CRITICAL()` with no
arguments, which ESP-IDF's non-SMP FreeRTOS does not provide.

## Checklist


FreeRTOS/Demo/Common/Minimal/QPeek.c
```diff
@@ -97,10 +100,30 @@ void vStartQueuePeekTasks( void )
...
+             * the queue while strict priority order decides who runs: each task
+             * peeks and then suspends itself before the next lower priority task
+             * is allowed to look.  With more than one core a woken lower priority
+             * task runs at the same time on the other core and sees a value from
+             * a neighbouring step of the cascade instead.  Confine the four tasks
+             * to one core so the cascade keeps the order it describes. */
+            vTaskCoreAffinitySet( xLowPriorityTask, ( UBaseType_t ) ( 1U << 1 ) );
+            vTaskCoreAffinitySet( xMediumPriorityTask, ( UBaseType_t ) ( 1U << 1 ) );
+            vTaskCoreAffinitySet( xHighPriorityTask, ( UBaseType_t ) ( 1U << 1 ) );
+            vTaskCoreAffinitySet( xHighestPriorityTask, ( UBaseType_t ) ( 1U << 1 ) );
```
chinglee-iot, line 125: Using core affinity forces these four tasks onto a single core so the single-core priority scheduler controls their execution order.

I have two suggestions regarding this change:

1. **Alternative Approach**: Could we use Task Notifications (xTaskNotifyGive / ulTaskNotifyTake) to explicitly sequence the tasks instead of relying on core affinity? This would allow the test to run reliably across multi-core systems without requiring affinity support.

2. **Compile Guard**: If we stick with core affinity, should we add an #else block with an #error directive? That way, if configNUMBER_OF_CORES > 1 but configUSE_CORE_AFFINITY or configRUN_MULTIPLE_PRIORITIES is disabled, the test fails at compile time with a clear explanation rather than silently skipping the affinity setup and failing at runtime.

FreeRTOS/Demo/Common/Minimal/EventGroupsDemo.c
```diff
@@ -91,6 +91,41 @@
...
+        for( ; ; )
+        {
+            eCurrent = eTaskGetState( xTask );
+
+            if( eCurrent == eExpected )
+            {
+                return eCurrent;
+            }
+
+            if( ( xTaskGetTickCount() - xStart ) > pdMS_TO_TICKS( 100 ) )
```
chinglee-iot, line 116:
```c
/* Maximum expected delay for a peer task state transition to complete 
 * under configRUN_MULTIPLE_PRIORITIES SMP execution. */
#define ebMULTIPLE_PRIORITIES_STATE_TRANSITION_MAX_TIME_MS    ( 100 )
```

Consider to define constant to indicate that the perpose of the 100ms waiting time.

FreeRTOS/Demo/Common/Minimal/IntQueue.c
```diff
@@ -608,11 +654,15 @@ static void prvLowerPriorityNormallyFullTask( void * pvParameters )
...
-            /* Should only succeed when the higher priority task is suspended */
-            if( eTaskGetState( xHighPriorityNormallyFullTask1 ) != eSuspended )
-            {
-                prvQueueAccessLogError( __LINE__ );
-            }
+            /* Should only succeed when the higher priority task is suspended.
+             * Racy on SMP (see prvLowerPriorityNormallyEmptyTask); guard to
+             * single core. */
+            #if ( configNUMBER_OF_CORES == 1 )
+                if( eTaskGetState( xHighPriorityNormallyFullTask1 ) != eSuspended )
```
chinglee-iot, line 661: This check is still valid when using a single priority configuration.

FreeRTOS/Demo/Common/Minimal/IntQueue.c
```diff
@@ -408,11 +447,18 @@ static void prvLowerPriorityNormallyEmptyTask( void * pvParameters )
...
-            {
-                prvQueueAccessLogError( __LINE__ );
-            }
+             * suspended.  This eTaskGetState() check is racy on SMP: the higher
+             * priority peer frees the queue (making this receive succeed) before
+             * it finishes suspending, so the lower priority task on the other
+             * core can observe the in-flight state instead of eSuspended.  Guard
+             * it to single core, matching the equivalent checks in recmutex.c and
+             * GenQTest.c; the queue-integrity checks keep the SMP coverage. */
+            #if ( configNUMBER_OF_CORES == 1 )
```
chinglee-iot, line 456: This check is still valid when using a single priority configuration.

FreeRTOS/Demo/Common/Minimal/TimerDemo.c
```diff
@@ -53,6 +53,80 @@
...
+ * Use a dedicated data-group critical section to atomically zero the
+ * callback counters at the start of each Test3 measurement so only firings
+ * that occur during the vTaskDelay are counted.  On granular-locks ports
+ * the plain taskENTER_CRITICAL() variant manipulates the port-shared user
+ * task lock whose state this test should not touch from a deep call chain;
+ * declare this file's own task and ISR spinlocks instead. */
+#if ( portUSING_GRANULAR_LOCKS == 1 ) && ( configNUMBER_OF_CORES > 1 )
+    static portSPINLOCK_TYPE xTimerDemoTaskLock = portINIT_SPINLOCK_STATIC;
+    static portSPINLOCK_TYPE xTimerDemoISRLock = portINIT_SPINLOCK_STATIC;
+#endif
```
chinglee-iot: `taskENTER_CRITICAL` and `taskEXIT_CRITICA`L are intended for application-level critical sections. Since user data-group critical sections are already used throughout this demo to serialize access, using them here to protect the timer demo data is consistent and appropriate. 

The other consideration is that original SMP with multiple priorities configuration can still run this demo without problem.

FreeRTOS/Demo/Common/Minimal/TimerDemo.c
```diff
@@ -601,29 +691,62 @@ static void prvTest6_CheckAutoReloadResetBehaviour( void )
...
+                configASSERT( xTestStatus );
+            }
+
+            if( ucAutoReloadTimerCounters[ configTIMER_QUEUE_LENGTH - 1 ] != ( uint8_t ) 0 )
+            {
+                xTestStatus = pdFAIL;
+                configASSERT( xTestStatus );
+            }
+
+            taskDATA_GROUP_EXIT_CRITICAL( &xTimerDemoTaskLock, &xTimerDemoISRLock );
```
chinglee-iot: Consider using standard taskENTER_CRITICAL / taskEXIT_CRITICAL for serialization here. Additionally, to resolve the code duplication in prvTest6, we could use conditional enter/exit macros rather than duplicating the entire test block. Also consider the multiple priorities SMP should be able to run the test.

FreeRTOS/Demo/Common/Minimal/TimerDemo.c
```diff
@@ -642,27 +765,77 @@ static void prvTest6_CheckAutoReloadResetBehaviour( void )
...
+            configASSERT( xTestStatus );
+        }
+
+        if( ucAutoReloadTimerCounters[ configTIMER_QUEUE_LENGTH - 1 ] == 0 )
+        {
+            xTestStatus = pdFAIL;
+            configASSERT( xTestStatus );
+        }
+
+        taskDATA_GROUP_EXIT_CRITICAL( &xTimerDemoTaskLock, &xTimerDemoISRLock );
```
chinglee-iot: Consider using standard taskENTER_CRITICAL / taskEXIT_CRITICAL for serialization here. Additionally, to resolve the code duplication in prvTest6, we could use conditional enter/exit macros rather than duplicating the entire test block. Also consider the multiple priorities SMP should be able to run the test.

FreeRTOS/Demo/Common/Minimal/StaticAllocation.c
```diff
@@ -48,6 +48,77 @@
...
+                if( ( xTaskGetTickCount() - xStart ) > pdMS_TO_TICKS( 100 ) )
+                {
+                    break;
+                }
+
+                vTaskDelay( 1 );
+            }
+        }
+
+        #define staticWAIT_FOR_TIMER_CALLBACKS( puxCount, uxExpected )    prvWaitForTimerCallbackCount( ( puxCount ), ( uxExpected ) )
```
chinglee-iot: The proposed solution in this commit addresses several critical SMP timing factors:

1. **Command Queue Propagation & Execution Shift**: Accounts for inter-core delays between queue posting on Core A and execution on Core B.

2. **Non-Deterministic Scheduling & Spinlock Contention**: Handles multi-core execution jitter where single-core priority guarantees no longer hold.

3. **Stack Lifespan Safety**: Ensures shared buffers remain safe in static storage if a caller frame unwinds.

However, relying on expected counter values as a synchronization mechanism is tailored primarily for test verification. It may not scale well as a general SMP design pattern for standard application code due to CPU polling overhead and latency.

For general application tasks that must resume strictly upon timer completion, using event-driven primitives like Task Notifications (xTaskNotifyGive / ulTaskNotifyTake) or Semaphores (xSemaphoreGive / xSemaphoreTake) provides a cleaner, more efficient synchronization mechanism across cores.

FreeRTOS/Demo/Common/Minimal/blocktim.c
```diff
@@ -114,8 +117,26 @@ void vCreateBlockTimeTasks( void )
...
         /* Create the two test tasks. */
-        xTaskCreate( vPrimaryBlockTimeTestTask, "BTest1", bktBLOCK_TIME_TASK_STACK_SIZE, NULL, bktPRIMARY_PRIORITY, NULL );
+        #if ( configNUMBER_OF_CORES > 1 ) && ( configUSE_CORE_AFFINITY == 1 ) && ( configRUN_MULTIPLE_PRIORITIES == 1 )
+            xTaskCreate( vPrimaryBlockTimeTestTask, "BTest1", bktBLOCK_TIME_TASK_STACK_SIZE, NULL, bktPRIMARY_PRIORITY, &xPrimary );
+        #else
+            xTaskCreate( vPrimaryBlockTimeTestTask, "BTest1", bktBLOCK_TIME_TASK_STACK_SIZE, NULL, bktPRIMARY_PRIORITY, NULL );
+        #endif
         xTaskCreate( vSecondaryBlockTimeTestTask, "BTest2", bktBLOCK_TIME_TASK_STACK_SIZE, NULL, bktSECONDARY_PRIORITY, &xSecondary );
+
+        #if ( configNUMBER_OF_CORES > 1 ) && ( configUSE_CORE_AFFINITY == 1 ) && ( configRUN_MULTIPLE_PRIORITIES == 1 )
```
chinglee-iot: We should add a check for that core affinity feature should be enabled if running with multiple priorities configuration .

FreeRTOS/Demo/Common/Minimal/recmutex.c
```diff
@@ -155,13 +209,15 @@ static void prvRecursiveMutexControllingTask( void * pvParameters )
...
-             * block time expires.  If the block time does expire then the error
-             * flag will be set here. */
-            if( xSemaphoreTakeRecursive( xMutex, recmu15ms_DELAY ) != pdPASS )
+             * the priority of this task.  The polling task can take longer than
+             * the block time to hand the mutex back (it holds the mutex while
+             * priority inheritance settles), so a single take can time out -
+             * retry until it succeeds so the number of takes always matches the
+             * number of gives below.  A genuine failure to ever obtain the mutex
+             * shows up as a stall that the check task detects. */
+            while( xSemaphoreTakeRecursive( xMutex, recmu15ms_DELAY ) != pdPASS )
```
chinglee-iot: Single core or single priority configuration doesn't need the loop. We should at least use bounded check here.

FreeRTOS/Demo/Common/Minimal/recmutex.c
```diff
@@ -228,25 +284,39 @@ static void prvRecursiveMutexBlockingTask( void * pvParameters )
...
+             * started its next cycle, leaving no instant at which this task can
+             * observe it suspended.  Check it only where that ordering is
+             * guaranteed, matching the peer state checks in the polling task
+             * below. */
+            #if !( ( configNUMBER_OF_CORES > 1 ) && ( configRUN_MULTIPLE_PRIORITIES == 1 ) )
+                if( prvWaitForSuspension( &xControllingIsSuspended ) != pdPASS )
                 {
                     xErrorOccurred = __LINE__;
                 }
+            #endif
```
chinglee-iot: Single core or single priority condition must be preserved here.

FreeRTOS/Demo/Common/Minimal/recmutex.c
```diff
@@ -255,11 +325,17 @@ static void prvRecursiveMutexBlockingTask( void * pvParameters )
             xErrorOccurred = __LINE__;
         }
 
-        /* The controlling and blocking tasks should be in lock step. */
-        if( uxControllingCycles != ( UBaseType_t ) ( uxBlockingCycles + 1 ) )
-        {
-            xErrorOccurred = __LINE__;
-        }
+        #if ( configNUMBER_OF_CORES == 1 )
```
chinglee-iot: Single-priority configurations still satisfy this check, so we should account for configRUN_MULTIPLE_PRIORITIES == 0 here as well.

FreeRTOS/Demo/Common/Minimal/recmutex.c
```diff
@@ -280,15 +356,33 @@ static void prvRecursiveMutexPollingTask( void * pvParameters )
...
+             * mutex in xSemaphoreGiveRecursive() before it sets its suspend
+             * flag and actually calls vTaskSuspend(), so the idle-priority
+             * polling task can acquire the freed mutex on the other core and
+             * read the peer task's state while it is still completing that
+             * give-then-suspend sequence.  eTaskGetState() then correctly
+             * reports the in-flight state (eRunning/eReady/eBlocked) rather than
+             * eSuspended.  Guard them to single core, matching the equivalent
+             * eTaskGetState() checks further below; the xXxxIsSuspended flag
+             * check that follows provides the SMP-safe coverage. */
+            #if ( INCLUDE_eTaskGetState == 1 ) && ( configNUMBER_OF_CORES == 1 )
```
chinglee-iot, line 285: Single-priority configurations still satisfy this check, so we should account for configRUN_MULTIPLE_PRIORITIES == 0 here as well.

FreeRTOS/Demo/Common/Minimal/recmutex.c
```diff
@@ -407,5 +527,10 @@ BaseType_t xAreRecursiveMutexTasksStillRunning( void )
         xReturn = pdPASS;
     }
 
+    /* Reset the sticky error flag so a single transient SMP timing slip
+     * does not permanently mark the test as failed.  Same pattern as
+     * AbortDelay / StreamBufferDemo. */
+    xErrorOccurred = pdFALSE;
```
chinglee-iot: These standard demo tests are designed to run long-term (e.g., 48 hours) to catch any latent issues. We shouldn't clear xErrorOccurred here; sticky errors need to be preserved so system check tasks can reliably detect and report a failure.

FreeRTOS/Demo/Common/Minimal/recmutex.c
```diff
@@ -280,15 +356,33 @@ static void prvRecursiveMutexPollingTask( void * pvParameters )
...
             }
             #endif /* INCLUDE_eTaskGetState */
 
-            /* Is the blocking task suspended? */
-            if( ( xBlockingIsSuspended != pdTRUE ) || ( xControllingIsSuspended != pdTRUE ) )
+            #if ( configNUMBER_OF_CORES > 1 ) && ( configRUN_MULTIPLE_PRIORITIES == 1 )
+
+                /* Multi-priority SMP: skip the flag check (racy) and always
+                 * proceed to the critical section. */
+                if( pdFALSE )
```
chinglee-iot: We should use a bounded wait on xBlockingIsSuspended and xControllingIsSuspended here rather than skipping the check under multi-priority SMP, maintaining consistent test behavior across all configurations.

FreeRTOS/Demo/Common/Minimal/IntSemTest.c
```diff
@@ -198,16 +225,36 @@ static void prvTakeAndGiveInTheSameOrder( void )
...
+     *     resumes (now at inherited priority) and reads the slave's state
+     *     while the slave is still physically executing the remaining few
+     *     instructions of xSemaphoreTake() before transitioning to eBlocked.
+     *   - eTaskGetState() correctly reports eRunning during that in-flight
+     *     transition.
+     * The priority inheritance check below (uxTaskPriorityGet(NULL) ==
+     * intsemSLAVE_PRIORITY) is the primary verification and is sufficient
+     * to prove the slave entered the blocking path: priority inheritance
+     * only happens inside xSemaphoreTake() when the mutex is contended. */
+    #if ( INCLUDE_eTaskGetState == 1 ) && ( configNUMBER_OF_CORES == 1 )
```
chinglee-iot, line 208: This condition is still satisfied with single priority configuration.

FreeRTOS/Demo/Common/Minimal/IntSemTest.c
```diff
@@ -300,16 +347,36 @@ static void prvTakeAndGiveInTheOppositeOrder( void )
...
```
chinglee-iot, line 310: This condition is still satisfied with single priority configuration.

FreeRTOS/Demo/Common/Minimal/QPeek.c
sudeep-mohanty: Dropped the core affinity approach; `QPeek.c` is now byte-identical to upstream. The demo leaves this test out under multiple priorities via `mainENABLE_ORDER_DEPENDENT_TESTS` in its own `main_full.c`, so the shared source is unchanged.

FreeRTOS/Demo/Common/Minimal/EventGroupsDemo.c
sudeep-mohanty: Removed; `EventGroupsDemo.c` is now byte-identical to upstream. The test is left out at demo level instead.

FreeRTOS/Demo/Common/Minimal/IntQueue.c
sudeep-mohanty: These checks don't hold under dual-core single priority either as `configRUN_MULTIPLE_PRIORITIES == 0` only stops tasks of different priorities running concurrently. However, in these tests, tasks of the same priority (due to priority inheritance) can run simultaneously on multiple cores. So we cannot reliably verify the state of a task running on another core. So, I've guarded the check to be active only for single-core configuration.

Changed in https://github.com/FreeRTOS/FreeRTOS/pull/1434/commits/0d9211318.

FreeRTOS/Demo/Common/Minimal/IntQueue.c
sudeep-mohanty: These checks don't hold under dual-core single priority either as `configRUN_MULTIPLE_PRIORITIES == 0` only stops tasks of different priorities running concurrently. However, in these tests, tasks of the same priority (due to priority inheritance) can run simultaneously on multiple cores. So we cannot reliably verify the state of a task running on another core. So, I've guarded the check to be active only for single-core configuration.

Here it is the receiver raising its own priority around each send rather than inheritance, with the same effect.

Changed in https://github.com/FreeRTOS/FreeRTOS/pull/1434/commits/0d9211318.

FreeRTOS/Demo/Common/Minimal/TimerDemo.c
sudeep-mohanty: Removed. `TimerDemo.c` now has a single change: the `prvTest1` queue-full sub-check keys off `xTaskGetSchedulerState()` so the demo adapts itself. Agreed on `taskENTER_CRITICAL` for application-level critical sections. See https://github.com/FreeRTOS/FreeRTOS/pull/1434/commits/b621815f8.

FreeRTOS/Demo/Common/Minimal/TimerDemo.c
sudeep-mohanty: Removed, along with the `prvTest6` duplication. See https://github.com/FreeRTOS/FreeRTOS/pull/1434/commits/b621815f8.

FreeRTOS/Demo/Common/Minimal/TimerDemo.c
sudeep-mohanty: Removed, along with the `prvTest6` duplication. See https://github.com/FreeRTOS/FreeRTOS/pull/1434/commits/b621815f8.

FreeRTOS/Demo/Common/Minimal/StaticAllocation.c
sudeep-mohanty: The counter polling is gone; those checks are now compiled out in configurations where the premise does not hold. Two changes remain, both lifetime fixes: `uxVariableToIncrement` is `static` because its address is passed as the timer ID and `xTimerDelete()` only queues the delete, and `xTCBBuffer` is `static` because `vTaskDelete()` does not tear down a task running on another core. See https://github.com/FreeRTOS/FreeRTOS/pull/1434/commits/c07db16bb.

FreeRTOS/Demo/Common/Minimal/blocktim.c
sudeep-mohanty: No longer applicable; `blocktim.c` is now byte-identical to upstream and the test is left out at demo level.

FreeRTOS/Demo/Common/Minimal/recmutex.c
sudeep-mohanty: The loop has been removed. See https://github.com/FreeRTOS/FreeRTOS/pull/1434/commits/f15b172da.

FreeRTOS/Demo/Common/Minimal/recmutex.c
sudeep-mohanty: Condition restored to upstream. See https://github.com/FreeRTOS/FreeRTOS/pull/1434/commits/f15b172da.

FreeRTOS/Demo/Common/Minimal/recmutex.c
sudeep-mohanty: These checks don't hold under dual-core single priority either as `configRUN_MULTIPLE_PRIORITIES == 0` only stops tasks of different priorities running concurrently. However, in these tests, tasks of the same priority (due to priority inheritance) can run simultaneously on multiple cores. So we cannot reliably verify the state of a task running on another core. So, I've guarded the check to be active only for single-core configuration.

Changed in https://github.com/FreeRTOS/FreeRTOS/pull/1434/commits/f15b172da.

FreeRTOS/Demo/Common/Minimal/recmutex.c
sudeep-mohanty: These checks don't hold under dual-core single priority either as `configRUN_MULTIPLE_PRIORITIES == 0` only stops tasks of different priorities running concurrently. However, in these tests, tasks of the same priority (due to priority inheritance) can run simultaneously on multiple cores. So we cannot reliably verify the state of a task running on another core. So, I've guarded the check to be active only for single-core configuration.

Changed in https://github.com/FreeRTOS/FreeRTOS/pull/1434/commits/f15b172da.

FreeRTOS/Demo/Common/Minimal/recmutex.c
sudeep-mohanty: Agreed - `xErrorOccurred` is no longer cleared. See https://github.com/FreeRTOS/FreeRTOS/pull/1434/commits/f15b172da.

FreeRTOS/Demo/Common/Minimal/recmutex.c
sudeep-mohanty: The skip has been removed and the check restored, with no bounded wait. See https://github.com/FreeRTOS/FreeRTOS/pull/1434/commits/f15b172da.

FreeRTOS/Demo/Common/Minimal/IntSemTest.c
sudeep-mohanty: These checks don't hold under dual-core single priority either as `configRUN_MULTIPLE_PRIORITIES == 0` only stops tasks of different priorities running concurrently. However, in these tests, tasks of the same priority (due to priority inheritance) can run simultaneously on multiple cores. So we cannot reliably verify the state of a task running on another core. So, I've guarded the check to be active only for single-core configuration.

Changed in https://github.com/FreeRTOS/FreeRTOS/pull/1434/commits/3ef18a237.

FreeRTOS/Demo/Common/Minimal/IntSemTest.c
sudeep-mohanty: These checks don't hold under dual-core single priority either as `configRUN_MULTIPLE_PRIORITIES == 0` only stops tasks of different priorities running concurrently. However, in these tests, tasks of the same priority (due to priority inheritance) can run simultaneously on multiple cores. So we cannot reliably verify the state of a task running on another core. So, I've guarded the check to be active only for single-core configuration.

Changed in https://github.com/FreeRTOS/FreeRTOS/pull/1434/commits/3ef18a237.

